PR #12, opened 2018-02-01
Mentions

adrianq:
Add terms and styling for mentions. Related to https://github.com/dhis2/d2-analysis/pull/54


edoardo:
In this app I always seem to get a 500 when trying to post an interpretation which uses a mention.
It works when the mention is removed.

adrianq:
@edoardo You are right. There was an error in the backend. I have just created [a PR that fixes this problem](https://github.com/dhis2/dhis2-core/pull/1900)

PR #19, opened 2018-06-20
Updating VI translations.

jason-p-pickering:


jason-p-pickering:
Will resync this. Some things do not look entirely correct to me.

PR #827, opened 2021-10-06
feat: populate metadata in store and use it to display item names in DataTable

edoardo:
* store metadata from the analytics response in the store
* use item names instead of ids in the `DataTable`

The `DataTable` looks much better now:

Screenshot 2021-10-06 at 15 01 22

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types'
 import React, { useState, useEffect } from 'react'
 import { connect } from 'react-redux'
 import { acClearCurrent, acSetCurrent } from '../actions/current'
+import { acAddMetadata } from '../actions/metadata'
 import { acSetUser } from '../actions/user'
 import {
     acClearVisualization,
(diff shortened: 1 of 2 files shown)


src/components/App.js
```diff
@@ -98,6 +100,24 @@ const App = ({
...
     }
 
+    const onResponseReceived = response => {
+        const metadata = Object.entries(response.metaData.items).reduce(
+            (obj, [id, item]) => {
+                obj[id] = {
+                    id,
+                    name: item.name || item.displayName,
+                    displayName: item.displayName,
+                    dimensionItemType: item.dimensionItemType, // TODO needed?
```
martinkrulltott, line 110: needed? 🤔

martinkrulltott (approved):

src/components/App.js
edoardo: I don't know, for now we don't use the metadata in the store, because the `DataTable` is rendered in the same component where the analytics response is fetched, so the metadata is read directly from there.
I didn't check how we use that in DV, but perhaps somewhere in the Layout is used?

PR #846, opened 2021-10-13
refactor: use custom hooks for fetching various data

edoardo:
Setting it to Draft as I'd like to use this PR for a discussion on how to use dataEngine/Query hooks in components and other custom hooks for splitting code.

One issue is that some components like `App.js` fetch various data and having all the query definitions and query handling in them can quickly cause the components to grow too big.

Dashboard app already uses some custom hook for fetching/storing data, so this approach is not new.
There are however different scenarios:
1. use of `lazy` and `refetch` like for `useVisualization`: the hook right now is just wrapping `useDataQuery`, basically the only advantage is that the query definition can be moved out of the component
2. common fetch functions imported from `analytics`: for these it would probably make more sense to export a hook directly from analytics but in this case it cannot do anything on the app Redux, but simply return some data.
3. “normal” queries that can be executed when the component is loaded
For all cases, error handling is also something I'm not sure how to solve, as hooks have some limitations.


edoardo:
Closing, we already implemented custom hooks for various requests, both in the app and in analytics.

PR #852, opened 2021-10-26
feat: download menu

edoardo:
Implements the download menu similar to the one in DV.

### TODO

- [ ] Advanced sub-menu?
- [ ] distinguish between event and enrollment
- [ ] more query parameters used in the old app, are they all needed/available?

### Screenshot
Screenshot 2021-10-26 at 14 21 06


HendrikThePendric:
Code from this PR was used to produce #874, which has been merged in already. So this PR can be closed.

PR #857, opened 2021-11-01
Chore/enable-deploy-previews-for-develop-branch

HendrikThePendric:
Implements PR deploy previews for the develop branch

### Key features

Live previews

### Description

The primary purpose of this branch is to work out why the netlify deployments are not happening currently. 

If the outcome is that this is not happening due to a lack of credentials, they should be added and this PR can be closed.

If we find another problem, we will fix it in this branch.


HendrikThePendric:
With this branch we've established that the problem is simply caused by missing credentials. We could still merge this in, but we don't need to.

varl:
Credentials have been added to the repo, along with the Netlify side of things that needed to be setup.

martinkrulltott (approved):

PR #859, opened 2021-11-03
fix(visualization): make table auto-width + add relativePeriodDate prop

HendrikThePendric:
### Key features

1. Makes the data-table have an automatic layout/width
2. Adds the relativePeriodDate prop

---

### Description

The datatable is now horizontally centered and has an automtic width. The auto-width feature is visible in the specs and I know from other apps I've recently worked on that this is a required functionality. However,if we want to keep showing the TitleBar before the datatable is ready, and the keep it seperate from the datatable, we can't left align the table and center-align the title above it. So in the end we need to decide between two options:
A. Put the title into the table, and accept this isn’t rendered when the table is loading. (we would have to find some way to not render the title when in a plugin)
B. Center the table and title as in the screenshot.

---

### TODO

-   [ ] Check with Joe about the best solution

---

### Known issues

-   [ ] Showing things in the horizontal center is not inline with the design specs

---

### Screenshots
Screenshot 2021-11-03 at 11 42 27

### DIFF
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -6,9 +6,9 @@ import {
     DataTableRow,
     DataTableCell,
     DataTableColumnHeader,
-    DataTableToolbar,
-    TableHead,
-    TableBody,
+    DataTableHead,
+    DataTableBody,
+    DataTableFoot,
     Pagination,
 } from '@dhis2/ui'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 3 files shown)


src/components/Visualization/Visualization.js
```diff
@@ -52,14 +56,19 @@ export const Visualization = ({ visualization, onResponseReceived }) => {
...
+        }
 
         const req = new analyticsEngine.request()
             .fromVisualization(visualization)
             .withProgram(visualization.program.id)
             .withStage(visualization.programStage.id)
             .withDisplayProperty('NAME') // TODO from settings ?!
             .withOutputType(visualization.outputType)
-            .withParameters({ completedOnly: visualization.completedOnly })
+            .withParameters(params)
```
edoardo, line 71: There is a specific method for the period date: `withRelativePeriodDate(relativePeriodDate)` although the generic `withParameters` also works.

edoardo (approved):

PR #860, opened 2021-11-05
feat: interpretations unit component

edoardo:
Implements [TECH-783](https://jira.dhis2.org/browse/TECH-783)

---

### Key features

1. component for rendering the interpretations list on the right sidebar
2. the component also implements the input for adding a new interpretation
3. it's possible to like an interpretation from the list
4. clicking on an interpretation "box" routes the app to a new URL containing the interpretation id, which opens the interpretation view (TECH-702)

---

### Description

See above :)

---

### TODO

-   [x] implement the "Write interpretation" feature

---

### Screenshots
Screenshot 2021-11-05 at 09 48 40

RichTextEditor:
Screenshot 2021-11-10 at 14 47 43
Screenshot 2021-11-10 at 14 46 51


HendrikThePendric (changes requested):
You have used a particular pattern for mutations, which probably works but is not really idiomatic. Apart from semantics there is also a potential functional issue: to ensure these mutation objects are static you are using `useMemo`, which I think is isn't guaranteed to keep referential equality, so this is probably a bad idea in the long run:
> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components.

See: https://reactjs.org/docs/hooks-reference.html#usememo

I think it might be easiest to illustrate the changes I have in mind by just doing a commit to this branch. I'll start preparing one now.

HendrikThePendric (approved):

src/components/DetailsPanel/DetailsPanel.js
```diff
@@ -12,7 +12,7 @@ export const DetailsPanel = ({ currentUser, visualization }) => (
     <div className={classes.panel}>
         <AboutVisualizationUnit type="eventReports" id={visualization.id} />
         <InterpretationsUnit
-            type="eventReports"
+            type="eventReport"
```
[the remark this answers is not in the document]
HendrikThePendric: Agreed

src/components/DetailsPanel/InterpretationsUnit/InterpretationForm.js
```diff
@@ -0,0 +1,77 @@
...
+    const inputPlaceholder = i18n.t('Write an interpretation')
+
+    return (
+        <div className={classes.container}>
+            <UserAvatar name={currentUser?.name} medium />
+            {showRichTextEditor ? (
+                <div className={classes.input}>
+                    <RichTextEditor
+                        inputPlaceholder={inputPlaceholder}
+                        onChange={value => setInterpretationText(value)}
```
HendrikThePendric, line 39: I think the change below would be functionally equivalent
```suggestion
                        onChange={setInterpretationText}
```

src/components/DetailsPanel/InterpretationsUnit/InterpretationList.js
```diff
@@ -28,27 +28,46 @@ export const InterpretationList = ({
...
+                                    const dateB = b.created
+
+                                    if (dateA < dateB) {
+                                        return -1
+                                    }
+                                    if (dateA > dateB) {
+                                        return 1
+                                    }
+                                    return 0
+                                })
```
HendrikThePendric, line 58: There is a lot going on here, not too much but getting close. I think extracting this function into a named helper called `sortByDateProp` is going to help clean this up enough.

src/components/DetailsPanel/InterpretationsUnit/Interpretation.js
```diff
@@ -0,0 +1,113 @@
...
+                        />
+                    </span>
+                </Tooltip>
+                <Tooltip closeDelay={200} content={i18n.t('Reply')}>
+                    <span onClick={() => onClick(interpretation.id)}>
+                        {interpretation.comments.length
+                            ? interpretation.comments.length
+                            : null}
+                        <IconReply16 color={colors.grey700} />
+                    </span>
```
HendrikThePendric, line 87: Semantically a `<button>` tag is probably a better fit here. With our supported browsers list we can use the CSS [`all` property](https://caniuse.com/css-all) to unset the browser styles (which effectively should make this look like a span?):
```css
button {
    all: unset;
    /* More styles */
}

HendrikThePendric (approved):
Left three minor comments, but LGTM

src/components/DetailsPanel/InterpretationsUnit/InterpretationList.js
edoardo: Sort function extracted.

src/components/DetailsPanel/InterpretationsUnit/InterpretationForm.js
edoardo: Fixed.

PR #861, opened 2021-11-11
feat: new org unit dimension (TECH-707)

martinkrulltott:
Implements [TECH-707](https://jira.dhis2.org/browse/TECH-707)

**Requires https://github.com/dhis2/analytics/pull/1081**

---

### Key features

1. Implement the org unit dimension from Analytics

---

### Description

Adds all the necessary parts to support the use of the org unit dimension from Analytics. Primarily the `DialogManager`, everything related to `rootOrgUnits` (now array instead of single object) and `parentGraphMap`.

To test this, open a stored AO made in the old app and click on the Org unit dimension in the layout.

The `AddToLayoutButton` was added but is obviously unused as there's no way to access a dimension that's not already in the layout (as there's no dimension panel yet).

Removed things related to `all` in the `TooltipContent`, as this isn't applicable to ER (as far as we know now).

Added `flex-grow: 1;` to the `StartScreen` as everything was pushed to the left (a side-effect of some other PR I suppose)

---

### TODO

-   [x] Should `renderPersistedContent` still be used in the `DialogManager`? Or can it be rendered like regular components now that the tree is fetched from `ui`?

---

### Screenshots

image


src/components/Dialogs/DialogManager.js
```diff
@@ -0,0 +1,242 @@
...
+import {
+    sGetUiItemsByDimension,
+    sGetUiActiveModalDialog,
+    sGetUiParentGraphMap,
+    sGetDimensionIdsFromLayout,
+} from '../../reducers/ui'
+import { AddToLayoutButton } from './AddToLayoutButton/AddToLayoutButton'
+
+export class DialogManager extends Component {
+    // TODO: Convert to functional component
```
edoardo, line 38: Is it possible to do this in this PR?

src/components/Dialogs/DialogManager.js
```diff
@@ -0,0 +1,242 @@
...
+
+        return null
+    }
+
+    renderDialogContent = () => {
+        const dimensionProps = {
+            onSelect: this.selectUiItems,
+        }
+
+        return <>{this.renderPersistedContent(dimensionProps)}</>
```
edoardo, line 137: Would this work?

```suggestion
        this.renderPersistedContent(dimensionProps)
```

edoardo (approved):

PR #864, opened 2021-11-17
fix: fixes and improvements to the RichTextEditor

edoardo:
### Key features

1. add `disabled` prop
2. implement a basic markdown handler

---

### Description

The `disabled` prop can be controlled by the loading state of a mutation, typically when you save the form where the rich text is sent in the payload.

The basic markdown handler implements a keypress listener to trigger bold and italic markdown when using a keyboard shortcut.
The handler inserts the markdown markers when keyboard shortcuts or the buttons in the toolbar are used and looks at the cursor position and at the text selection (if any).
The handler is not to be considered complete and put on hold for now until a decision is made about the use of a 3rd party library for the rich text editor/parser.


HendrikThePendric (approved):
Excellent!

src/components/DetailsPanel/InterpretationsUnit/RichTextEditor/RichTextEditor.js
```diff
@@ -13,53 +12,66 @@ import {
...
+} from './markdownHandler'
+import {
+    mainClasses,
+    toolbarClasses,
+    emojisPopoverClasses,
+} from './styles/RichTextEditor.style.js'
 
-const emoticonsButtonRef = createRef()
+const emojisButtonRef = createRef()
+const textareaRef = createRef()
```
HendrikThePendric, line 32: You should use `useRef` in the component function body. This creates a single ref for the module (~file) so this could get very problematic if there are multiple RickTextEditors in view at once.

HendrikThePendric (changes requested):
Sorry, I missed one part when I previously reviewed. My code comment is quite critical, so I've changed the review status to "request changes".

HendrikThePendric (approved):

PR #865, opened 2021-11-24
feat: edit and delete interpretations (TECH-824)

HendrikThePendric:
Implements [TECH-824](https://jira.dhis2.org/browse/TECH-824)

### Key features

1. Interpretations can be edited and deleted in the details pane
1. Interpretations can be edited and deleted in the interpretations modal
1. The modal keeps track of relevant changes to the interpretation. If any are detected, it will ensure the side-pane is updated when the modal closes 

### Description

Editing an interpretation is now possible from both the details pane as well as the interpretation modal. During development I encountered a challenge: 

Quite a few possible actions in the modal will update an interpretation but since the modal and the side-pane are completely separate components without shared state, the interpretation in the side pane might become stale. Solving this using a shared state was not a good option, because that would mean the app would have to maintain this shared state and that is not good since these are meant to become self-contained reusable components.

So instead of solving this by lifting state up, I have added a `handleInterpretationUpdate` callback prop to the modal, and have added an "imperative handle" to the interpretations unit. This is not the most idiomatic react pattern, but I think in this case the most appropriate choice.

### Screenshots

Screenshot 2021-11-24 at 16 19 37
Screenshot 2021-11-24 at 16 20 08


src/components/Interpretations/InterpretationsUnit/InterpretationsUnit.js
```diff
@@ -31,96 +37,151 @@ const interpretationsQuery = {
...
-                }
-            `}</style>
-        </div>
-    )
-}
+        )
+    }
+)
+
+InterpretationsUnit.displayName = 'InterpretationsUnit'
```
edoardo, line 184: Is this to avoid the linting error when importing default?

src/components/Interpretations/InterpretationsUnit/InterpretationList.js
```diff
@@ -59,8 +59,10 @@ export const InterpretationList = ({
                                         key={interpretation.id}
                                         interpretation={interpretation}
                                         currentUser={currentUser}
-                                        onClick={onInterpretationClick}
-                                        refresh={refresh}
+                                        reply={onInterpretationClick}
```
edoardo, line 62: Perhaps `reply` is not the most correct term, the click opens the modal where you it's possible to reply, but is not necessarily what will happen, ie. you can edit the interpretation/comment, simply read the comments or do nothing.

src/components/Interpretations/common/Interpretation/InterpretationDeleteButton.js
```diff
@@ -0,0 +1,34 @@
...
+        />
+    )
+}
+
+InterpretationDeleteButton.propTypes = {
+    id: PropTypes.string.isRequired,
+    onComplete: PropTypes.func.isRequired,
+}
+
+export { InterpretationDeleteButton }
```
edoardo, line 34: Is this any different than directly export the const above?

edoardo (approved):

src/components/Interpretations/InterpretationModal/InterpretationModal.js
```diff
@@ -34,10 +36,11 @@ const query = {
         id: ({ id }) => id,
         params: {
             fields: [
+                'access',
                 'id',
                 'text',
                 'created',
-                'createdBy[id,displayName]',
+                'user[id,displayName]',
```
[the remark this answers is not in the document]
HendrikThePendric: Yeah, consistency and reusability. The `InterpretationsUnit` uses `user[id,displayName]` and it helps if the property is the same when rendering the `UserAvatar` in the `Interpretation`, which is a shared component.

src/components/Interpretations/InterpretationsUnit/InterpretationsUnit.js
HendrikThePendric: All components that are wrapped in `forwardRef` should have a `displayName`, otherwise they show up in React DevTools as `Component`. In most DHIS2 projects not adding the `displayName` will result in an ESLint error, I'm not sure if I got one here, or if I simply added it out of habbit. In any case, it will help when debugging stuff in DevTools and will also help in Jest tests.

src/components/Interpretations/common/Interpretation/InterpretationDeleteButton.js
HendrikThePendric: No it's the same

edoardo (approved):

PR #875, opened 2021-11-30
feat: numeric condition (TECH-788)

martinkrulltott:

**Requires https://github.com/dhis2/analytics/pull/1097**

---

### Key features

1. Initial implementation of the conditions modal
2. Adds support for numeric conditions
3. Fetches legendSets for future use (for implementing range-sets)
4. Adds fixed dimensions metadata by default

---

### Description

This is the first implementation of the conditions modal, which assumes that all conditions are numeric (while waiting for `dataType` to be returned by the `/eventReports` API). 

Conditions are fetched as e.g. `GT:15` from the backend (`current`, `visualization`), is stored as `{[dimensionId]: condition}` in `ui` and is handled as an array of strings by the `ConditionsManager`. E.g. multiple conditions are parsed from `GT:15:LT:25` to `['GT:15', 'LT:25']` and back again for the handling in the `ConditionsManager` - this enables "incomplete" conditions (e.g. selecting just the operator 'GT' before providing a value `15`.
The `ConditionsManager` then passes this back to the store in the single-string format upon closing the conditions modal (note that this is a new pattern for modals, which otherwise usually has a direct connection with the store).

The `DialogManager` was refactored, from being the single handler of all dimension modals, to instead being just a router that directs to either the `FixedDimension` (org unit, periods etc) or the `ConditionsManager` (data items) depending on the dimension type. This was done to cater for the new store-on-close pattern described above (which needs direct access to the `Modal`'s `onClose` prop.

LegendSets (which are the same thing as range sets) are being fetched, as it will be used to implement range-sets in an upcoming PR (however currently pending backend support). They're fetched on app start and stored as `legendSets` on the root level of the store, to be used as a lookup table (similar to `metadata`).

The default `metadata` was missing the fixed dimensions (pe, ou), which made consumers rely on `dimensions` for this instead, the fixed dimension are now added by default.

Note: `dimensionId` is being used by the `ConditionsManager`'s `mapStateToProps`, but not by the component itself, so the eslint error was be turned off to reflect this.

---

### TODO

-   [x] Prevent "incomplete" conditions from being saved to the store
-   [ ] Implement range-sets for numeric conditions _(separate PR)_

---

### Known issue

-   [ ] Non-numeric conditions won't work and will throw an error (will be implemented in a separate PR)

---

### Screenshots

_no conditions_
image

_add a condition_
image

_one condition in use_
image

_two conditions in use_
image

_condition indicator on the chip_
image


src/reducers/ui.js
```diff
@@ -71,11 +71,6 @@ const getPreselectedUi = options => {
 
 export default (state = EMPTY_UI, action) => {
     switch (action.type) {
-        case SET_UI: {
```
martinkrulltott, line 74: Removed, as it was unused + setting ui directly is a bit of an anti-pattern we want to avoid anyway

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -0,0 +1,165 @@
...
+
+    const removeCondition = id =>
+        setConditionsList(conditionsList.filter((_, index) => index !== id))
+
+    const setCondition = (id, value) =>
+        setConditionsList(
+            conditionsList.map((condition, index) =>
+                index === id ? value : condition
+            )
+        )
```
HendrikThePendric, line 43: Perhaps consider renaming `id` to `conditionIndex` or sth, since all of this seems index based.

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -0,0 +1,165 @@
...
+                    content={renderModalContent()}
+                    dataTest={'dialog-manager-modal'}
+                    isInLayout={isInLayout}
+                    onClose={closeModal}
+                    onUpdate={primaryOnClick}
+                    title={renderModalTitle()}
+                />
+            )}
+        </>
+    )
```
HendrikThePendric, line 137: Or without a fragment:
```suggestion
    return dimension ? (
        <DimensionModal
            content={renderModalContent()}
            dataTest={'dialog-manager-modal'}
            isInLayout={isInLayout}
            onClose={closeModal}
            onUpdate={primaryOnClick}
            title={renderModalTitle()}
        />
    ) : null
```

src/components/Dialogs/Conditions/NumericCondition.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const NumericCondition = ({ condition, onChange, onRemove }) => {
+    const setOperator = input => {
+        if (input.includes(NULL_VALUE)) {
+            onChange(`${input}`)
+        } else {
+            onChange(`${input}:${value || ''}`)
+        }
+    }
```
HendrikThePendric, line 35: Should this function accept two args: `input` and `value`?
I see `value` is being used in the `else` clause but I don't see where its value would come from.

src/components/Dialogs/Conditions/NumericCondition.js
```diff
@@ -0,0 +1,87 @@
...
+    const setValue = input => onChange(`${operator}:${input || ''}`)
```
HendrikThePendric, line 36: Same as previous comment. Where is `operator` coming from?

src/components/Dialogs/DimensionModal.js
```diff
@@ -0,0 +1,68 @@
...
+
+const DimensionModal = ({
+    content,
+    dataTest,
+    isInLayout,
+    onClose,
+    onUpdate,
+    title,
+}) => {
+    return (
```
HendrikThePendric, line 22: This could be an implicit return, i.e.
```js
}) => (
```

src/components/Dialogs/FixedDimension.js
```diff
@@ -0,0 +1,173 @@
...
+            // TODO: case DIMENSION_ID_PERIOD:
+        }
+    }
+
+    const primaryOnClick = () => {
+        onUpdate()
+        closeModal()
+    }
+
+    const renderModalTitle = () => dimension.name
```
HendrikThePendric, line 120: You can just pass the string to `title`

src/components/Dialogs/DimensionModal.js
```diff
@@ -0,0 +1,68 @@
...
+        <Modal onClose={onClose} dataTest={`${dataTest}`} position="top" large>
+            <ModalTitle dataTest={`${dataTest}-title}`}>{title}</ModalTitle>
+            <ModalContent dataTest={`${dataTest}-content`}>
+                {content}
```
HendrikThePendric, line 26: I think we should consider using the `children` prop for populating the modal content. This will enable us to move away from the `renderModalContent` helpers and simply insert them into the `DimensionsModal`.

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -0,0 +1,165 @@
...
+                    dataTest={'conditions-manager-add-condition'}
+                    className={classes.addConditionButton}
+                >
+                    {conditionsList.length
+                        ? i18n.t('Add another condition')
+                        : i18n.t('Add a condition')}
+                </Button>
+            </div>
+        </>
+    )
```
HendrikThePendric, line 105: See my other comment about leveraging the `children` prop in the `DimensionModal` for this.

HendrikThePendric (commented):
I've left a few suggestions. The point I feel strongest about is replacing the `content` prop in the `DimensionDialog` by a `children` prop, so we can simply insert content into it.

I will leave the approving/requesting changes for team members who have more experience in this team than I have.

src/components/Dialogs/Conditions/NumericCondition.js
martinkrulltott: `operator` and `value` are global consts for the `NumericCondition` and are set here https://github.com/dhis2/event-reports-app/blob/67d6a29aca8d2037e68f39105d5d67a451485e7d/src/components/Dialogs/Conditions/NumericCondition.js#L38
Maybe this should be renamed or the whole section be reordered to above this fn?

src/components/Dialogs/Conditions/NumericCondition.js
martinkrulltott: See comment above ⏫

src/components/Dialogs/FixedDimension.js
martinkrulltott: yeah, good spot, too much copy pasting on my part 😄

src/components/Dialogs/DimensionModal.js
martinkrulltott: Good point! Refactoring this now!

src/components/Dialogs/Conditions/NumericCondition.js
HendrikThePendric: I had to think a for a minute why this code actually runs, because the functions that use these variables are declared before these variables themselves. So yeah, I think it would help (me) if you moved this variable declaration to above line 29.

src/components/Dialogs/Conditions/NumericCondition.js
HendrikThePendric: Same as above

HendrikThePendric (approved):
One small suggestion remains, but this looks good to me.

PR #878, opened 2021-12-06
TECH-825-modal-open-content-shift

HendrikThePendric:
Implements [TECH-825](https://jira.dhis2.org/browse/TECH-825)

---

### Key features

1. Use SWR in the `Visualization` component to avoid content shifts when interacting with the visualisation in the modal
2. Extract the `Visualization` component's data fetching logic into a hook
3. Show loading errors in the `Visualisation` component
4. Fix content shift in the modal on first load
5. Switch the global `isVisualizationLoading` state to `false` when closing the modal

---

### Description

I've reworked the Visualization component a bit. The code has been refactored so that the data logic is extracted into a hook. Functionality-wise I've also added support for rendering errors, but the biggest change is that this component now uses the SWR model. **S**tale **W**hile **R**evalidate means that we keep the old data in place while we are fetching the new data, and to indicate the data the user is currently viewing is stale, we show a loading indicator _on top of_ the stale data.

After applying SWR the content shifts in the modal are fairly limited. The user can sort data, navigate through pages, and the size of the table will remain relatively constant. We can probably do even more to keep the modal completely still, but this is already a vast improvement.

The above fixes the content shifts _once the modal has rendered_ and the user starts interacting with it. But that was only part of the problem. The modal worst content shift was happening when the modal was opening along with the app (i.e. when there is an `interpretationId` in the URL query params). The root cause of this content shift was caused by the fact that I was using the global `isVisualizationLoading` redux state property to determine when to show the modal. However, since this global variable is used by the main visualisation as well, this wasn't working properly: the main visualisation would be done loading when the visualisation in the modal wasn't done yet and this would cause the modal to render without the visualisation. To remedy this, the modal now has its own internal loading state.

~Even though the interpretation modal itself is now fully independent of the global loading state, opening it is still causing some changes in the global state via the history listener in `App.js`. It could be that we want to completely disable this behaviour later on, but I decided to keep it for now, because it also comes with some advantages: I am not showing the modal until all of its parts are done rendering, so this can take some time. The current behaviour ensures that the main UI goes into a loading state while the users waits for the modal to show, so that's nice. The downside of the current behaviour was that the main UI would keep that loading state when closing the modal. But it turns out this was [extremely easy to fix](https://github.com/dhis2/event-reports-app/blob/026275f8793aea92bca56801797de75a3f990f05/src/components/InterpretationModal/InterpretationModal.js#L21-L24), so I decided to just implement the quick fix for now. Refactoring the state might still be useful, but we can defer that until when all the moving parts are in place.~ This is no longer accurate, I've addressed this in baf705f, 3454408, 3f64840 and 4d83236. The modal now has it's own loader UI and is completely independent from the global state 🚀 .

----

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-12-01T12:53:09.293Z\n"
-"PO-Revision-Date: 2021-12-01T12:53:09.293Z\n"
+"POT-Creation-Date: 2021-12-06T15:51:12.920Z\n"
+"PO-Revision-Date: 2021-12-06T15:51:12.920Z\n"
 
 msgid "Add to {{axisName}}"
 msgstr "Add to {{axisName}}"
@@ -205,6 +205,16 @@ msgstr "TEXT"
 
 msgid "Your most viewed TEXT"
 msgstr "Your most viewed TEXT"
+
+msgid "Could not load visualization"
+msgstr "Could not load visualization"
+
+msgid ""
+"The visualization couldn’t be displayed. Try again or contact your system "
+"administrator."
+msgstr ""
+"The visualization couldn’t be displayed. Try again or contact your system "
+"administrator."
 
 msgid "Cases per page"
 msgstr "Cases per page"
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -174,15 +174,16 @@ const App = ({
             const isSaving = location.state?.isSaving
             const isOpening = location.state?.isOpening
             const isResetting = location.state?.isResetting
+            const isModalOpening = location.state?.isModalOpening
+            const isModalClosing = location.state?.isModalClosing
+            const isValidLocationChange =
+                previousLocation !== location.pathname &&
+                !isModalOpening &&
+                !isModalClosing
 
             // TODO navigation confirm dialog
 
-            if (
-                isSaving ||
-                isOpening ||
-                isResetting ||
-                previousLocation !== location.pathname
-            ) {
+            if (isSaving || isOpening || isResetting || isValidLocationChange) {
                 loadVisualization(location)
             }
         })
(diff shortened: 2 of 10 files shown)


HendrikThePendric:
<details>
<summary>Here's some annotated `console.log` output:</summary>
<p>
```
++++ Visualization ++++ 
hasData:  false 
loading:  true 
fetching:  true 
inModal:  false                 // FIRST THE MAIN VISUALIZATION STARTS LOADING
------------
++++ InterpretationModal ++++   // BUT THE MODAL IS OPEN TOO AND WILL START LOADING ITS VISUALIZATION
loading:  false 
isVisualizationLoading:  true 
---------------
++++ InterpretationModal ++++ 
loading:  true 
isVisualizationLoading:  true 
---------------
++++ InterpretationModal ++++ 
loading:  true 
isVisualizationLoading:  true 
---------------
++++ Visualization ++++ 
hasData:  false 
loading:  true 
fetching:  true 
inModal:  false 
------------
++++ InterpretationModal ++++ 
loading:  true 
isVisualizationLoading:  true 
---------------
++++ Visualization ++++ 
hasData:  true 
loading:  true 
fetching:  true 
inModal:  false 
------------
++++ Visualization ++++ 
hasData:  true 
loading:  true 
fetching:  true 
inModal:  false 
------------
++++ InterpretationModal ++++ 
loading:  true 
isVisualizationLoading:  false 
---------------
++++ Visualization ++++ 
hasData:  true 
loading:  false 
fetching:  true 
inModal:  false 
------------
++++ Visualization ++++ 
hasData:  true 
loading:  false 
fetching:  false 
inModal:  false 
------------
++++ InterpretationModal ++++ 
loading:  false 
isVisualizationLoading:  false  // FALSE HERE ACTUALLY REFERS TO THE MAIN VISUALISATION LOADING STATE
---------------
++++ Visualization ++++ 
hasData:  false                 // BUT THE MODAL DOES NOT HAVE DATA YET. THIS TRIGGERS THE CONTENT SHIFT
loading:  true 
fetching:  true 
inModal:  true 
------------
++++ Visualization ++++ 
hasData:  true 
loading:  true 
fetching:  true 
inModal:  true 
------------
++++ Visualization ++++ 
hasData:  true 
loading:  false 
fetching:  true 
inModal:  true 
------------
++++ Visualization ++++ 
hasData:  true 
loading:  false 
fetching:  false 
inModal:  true 
------------
```
</p>
</details>

edoardo (approved):

PR #880, opened 2021-12-14
feat: alphanumeric conditions (TECH-812)

martinkrulltott:
Implements [TECH-812](https://jira.dhis2.org/browse/TECH-812)

---

### Key features

1. Implement alphanumeric conditions

---

### Description

Based on the recent implementation of numeric conditions. Plus some added logic for the case sensitive (CS), as CS operators are regular operators but needs to be joined with the non-CS counterparts (`SingleSelect` only allows passing in a `selected` that's actually in the list of children, so had to map the CS and non-CS together).

---

### Known issues

-   [ ] Analytics requests using most operators (e.g. `ILIKE`) doesn't work (lack of backend support?)

---

### Screenshots

image

### DIFF
--- a/src/components/Dialogs/Conditions/AlphanumericCondition.js
+++ b/src/components/Dialogs/Conditions/AlphanumericCondition.js
@@ -1,0 +1,99 @@
+import i18n from '@dhis2/d2-i18n'
+import { SingleSelectField, SingleSelectOption, Button, Input } from '@dhis2/ui'
+import PropTypes from 'prop-types'
+import React from 'react'
+import classes from './styles/Condition.module.css'
+
+const NULL_VALUE = 'NV'
+export const OPERATOR_EQUAL = 'EQ'
+export const OPERATOR_NOT_EQUAL = '!EQ'
+export const OPERATOR_CONTAINS = 'LIKE'
+export const OPERATOR_NOT_CONTAINS = '!LIKE'
+export const OPERATOR_EMPTY = `EQ:${NULL_VALUE}`
+export const OPERATOR_NOT_EMPTY = `NE:${NULL_VALUE}`
+
+const operators = {
+    [OPERATOR_EQUAL]: i18n.t('exactly'),
+    [OPERATOR_NOT_EQUAL]: i18n.t('is not'),
+    [OPERATOR_CONTAINS]: i18n.t('contains'),
+    [OPERATOR_NOT_CONTAINS]: i18n.t('does not contain'),
+    [OPERATOR_EMPTY]: i18n.t('is empty / null'),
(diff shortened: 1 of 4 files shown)


src/components/Dialogs/Conditions/AlphanumericCondition.js
```diff
@@ -0,0 +1,171 @@
...
+            case OPERATOR_NOT_CONTAINS:
+            case OPERATOR_NOT_CONTAINS_CASE_SENSITIVE:
+                setOperator(
+                    state
+                        ? OPERATOR_NOT_CONTAINS_CASE_SENSITIVE
+                        : OPERATOR_NOT_CONTAINS
+                )
+                break
+        }
+    }
```
HendrikThePendric, line 93: I think this would work too. It's less readable, but shorter and would still work when new operators are added. Not sure sure how relevant that is though.
```suggestion
    const suffix = '_CASE_SENSITIVE'
    
    const toggleCaseSensitive = (isCaseSensitive) => {
        const newOperator = isCaseSensitive
            ? operator + suffix
            : operator.replace(suffix, '')

        if ([newOperator]) {
            setOperator(newOperator)
        }
    }
```

src/components/Dialogs/Conditions/AlphanumericCondition.js
```diff
@@ -0,0 +1,171 @@
...
+            case OPERATOR_NOT_EQUAL_CASE_SENSITIVE:
+                return OPERATOR_NOT_EQUAL
+            case OPERATOR_CONTAINS_CASE_SENSITIVE:
+                return OPERATOR_CONTAINS
+            case OPERATOR_NOT_CONTAINS_CASE_SENSITIVE:
+                return OPERATOR_NOT_CONTAINS
+            default:
+                return operator
+        }
+    }
```
HendrikThePendric, line 108: I think this would work too. If the case sensitive suffix is not present, nothing is replaced.
```suggestion
    const useGenericOperator = (operator) => operator.replace(suffix, '')
```

HendrikThePendric (approved):
Looks good. I left a few comments.

src/components/Dialogs/Conditions/AlphanumericCondition.js
martinkrulltott: Thanks for the suggestion. This got me thinking... If we're going down the `const suffix = '_CASE_SENSITIVE'` route, we might as well just treat the case sensitive part as a suffix (however as a prefix) in the const itself and just use the regular operators and append the ´I´ when needed. I'll see what I can put together...

HendrikThePendric (approved):
Good solution!

PR #881, opened 2021-12-14
feat: add main sidebar and input accessory panel (TECH-872 / TECH-875)

HendrikThePendric:
Implements:
-  [TECH-872](https://jira.dhis2.org/browse/TECH-872)
-  [TECH-875](https://jira.dhis2.org/browse/TECH-875)

---

### Key features

1. Toggling animated panels
2. Working input accessory panel

---

#### Panels
- Have a nice slide-in/out animation
- When details panel opens, accessory panel closes and vice versa
- When selecting an item in the main sidebar the accessory panel opens and the corresponding content will be displayed in the accessory panel
- When clicking a selected item (accessory panel will be open) in the main sidebar, the accessory panel closes
- When clicking a non-selected item (accessory panel will be open) in the main sidebar the accessory panel stays open and the content in the accessory panel will be updated.

#### Input panel
- A selection can be made
- Selected value is stored in the main panel, so it can be passed to the program dimensions panel, where it will be used as a filter
- Selection is also displayed in the main side bar, i.e. `Choose an input` (no selection) or `Input: event/enrollment`

#### Assorted fixes
- Fix double loader/spinner on interpretations unit
- Add some whitespace left and right of the visualisation, to match the design specs (without this white space the main/accessory panel look like they have a double border)

#### Screenshot with open accessory panel
Screenshot 2021-12-15 at 09 27 25

#### Names of various sections
names

#### Animations in action
animations


src/actions/ui.js
```diff
@@ -7,14 +7,22 @@ import {
...
     ADD_UI_PARENT_GRAPH_MAP,
     SET_UI_CONDITIONS,
     sGetUiConditions,
+    SET_UI_INPUT,
 } from '../reducers/ui.js'
 
+export const acSetUiInput = (value) => ({
+    type: SET_UI_INPUT,
+    // Store input as an object but allow passing a string for simple input types
+    value: typeof value === 'string' ? { type: value } : value,
```
martinkrulltott, line 23: Unsure about this pattern. Its a clever solution and I get the idea, but to me its a bit over-engineered for this rather simple use-case. Would prefer if the consumer always has to provide the same params regardless of input type (i.e. `{ type: value }`), just for clarity and reduced amount of head scratching down the line 😄

HendrikThePendric: Yeah, I just figured we only need to store this as an object for the tracked entity only, but on balance, just sending the object when dispatching the action is more explicit and future proof.

HendrikThePendric: Addressed in d3aba95

src/components/MainSidebar/MainSidebar.js
```diff
@@ -0,0 +1,80 @@
...
+                                  nsSeparator: '^^',
+                              })
+                            : i18n.t('Choose an input')
+                    }
+                    onClick={() => onClick(IDS.INPUT)}
+                    selected={open && dimensionId === IDS.INPUT}
+                />
+                <MenuItem
+                    icon={<IconFolder16 />}
+                    label="Program dimensions"
```
edoardo, line 57:
```suggestion
                    label={i18n.t("Program dimensions")}
```

src/components/MainSidebar/MainSidebar.js
```diff
@@ -0,0 +1,80 @@
...
+                    onClick={() => onClick(IDS.PROGRAM)}
+                    selected={open && dimensionId === IDS.PROGRAM}
+                />
+                <MenuItem
+                    icon={<IconFolder16 />}
+                    label="Your dimensions"
```
edoardo, line 63:
```suggestion
                    label={i18n.t("Your dimensions")}
```

edoardo (approved):

PR #885, opened 2021-12-21
feat: date, datetime, time conditions (TECH-817)

martinkrulltott:
Implements [TECH-817](https://jira.dhis2.org/browse/TECH-817)

---

### Key features

1. Implement date, datetime, time conditions
2. Refactor all other condition types to follow the same prop signature

---

### Description

_feat_: Generally follows the pattern used for the numeric, alphanumeric and boolean conditions to implement date / datetime / time conditions. Main exception is that instead of having a single export with props to toggle various parts, separate exports for each sub-type are used.

_refactor_: All the previous condition types have all been refactored to use this new format with separately exported components rather than proprietary props unique to each condition type. This results in all conditions now use a much simpler unified prop signature instead: `condition`, `onChange` and the optional `onRemove`.

_refactor_: All condition operators and consts were moved to `/modules/conditions.js` instead of redundantly being defined in each condition file.

_note_: As `:` is reserved for splitting `condition:value` in the backend, its assumed that this format won't valid to use for the hour/minute divider (e.g. `15:30`), so instead the backend format is parsed to `:` before being passed on to the `ui` component (that expects `:`). However, as the preferred backend format / divider character is unknown, `.` (a single dot) is used in its place temporarily. This of course means that the `datetime` and `time` condition types won't work with the API for now.

---

### TODO

-   [ ] Change the `API_TIME_DIVIDER` to the proper character used by the backend

---

### Screenshots

_date_

image

_time_

image

_datetime_

image

### DIFF
--- a/src/components/Dialogs/Conditions/ConditionsManager.js
+++ b/src/components/Dialogs/Conditions/ConditionsManager.js
@@ -71,7 +71,8 @@ const ConditionsManager = ({
     onClose,
     setConditionsByDimension,
 }) => {
-    const dimensionType = DIMENSION_TYPE_TIME // TODO: Should be returned by the backend, e.g. NUMBER, INTEGER, PERCENTAGE
+    // const dimensionType = dimension.type
+    const dimensionType = DIMENSION_TYPE_DATE // TODO: Should be returned by the backend, e.g. NUMBER, INTEGER, PERCENTAGE
 
     const [conditionsList, setConditionsList] = useState(
         (conditions.condition?.length &&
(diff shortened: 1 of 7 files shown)


src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -162,42 +175,95 @@ const ConditionsManager = ({
...
+                            <TrueOnlyCondition
+                                condition={condition}
+                                onChange={(value) => setCondition(index, value)}
+                            />
+                        </div>
+                    )
+                )
+            }
+            case DIMENSION_TYPE_DATE: {
+                return conditionsList.map((condition, index) => (
```
edoardo, line 232: Should all check for `conditionsList.length` like the one above?

edoardo (approved):

src/components/Dialogs/Conditions/ConditionsManager.js
martinkrulltott: Actually, that's a copy paste error. Neither need to check for `conditionsList.length`, that was used for the numeric types to switch between either `conditionsList` or `legendSet`. Remove the redundant checks

PR #887, opened 2021-12-23
feat: optionset condition (TECH-816)

martinkrulltott:
Implements [TECH-816](https://jira.dhis2.org/browse/TECH-816)

---

### Key features

1. Adds option set conditions (`Transfer` component)
2. Fetches `valueType` and `optionSet` props from the `eventReports` request and stores in `metadata`

---

### Description

When a visualization is loaded (i.e. on open), the `valueType` and `optionSet` metadata for each dimension are stored in the `metadata` section of the Redux store. This is then used to first determine which condition type to use (based on `valueType`) and then, if the option set type is detected (type `TEXT` + having an `optionSet`), the new `OptionSetCondition` is displayed. 
The actual `options` are fetched in a separate request to populate a `Transfer` component with selected (from the conditions) and available options (from the request).

Most of the code in `OptionSetCondition.js` was borrowed from / influenced by similar implementations of `Transfer` in Analytics ([example](https://github.com/dhis2/analytics/blob/master/src/components/DynamicDimension/ItemSelector.js)).

The `metadata` reducer was changed from overwriting objects to extending them instead.
e.g. `x: { prop: value }` + `x: { prop2: value2 }` = `x: { prop: value, prop2: value2 }`

`OPERATOR_IN` was added to further consolidate / centralise the operators.

The `metadata` middleware was extended to be able to handle either a single object or an array of objects.

---

### TODO

-   [ ] `useDebounce` and `useDidUpdateEffect` were copied from Analytics. Should these be exported from Analytics instead?

---

### Known issues

-   [ ] Metadata for conditions that are in use but aren't in the scope of the loaded options are missing their name and thus show up as empty lines (see screenshot below)

---

### Screenshots

_bug_

image


src/components/Dialogs/Conditions/OptionSetCondition.js
```diff
@@ -0,0 +1,205 @@
...
+        searchTerm: '',
+        options: [],
+        loading: true,
+        nextPage: 1,
+    })
+    const dataEngine = useDataEngine()
+    const setSearchTerm = (searchTerm) =>
+        setState((state) => ({ ...state, searchTerm }))
+    const debouncedSearchTerm = useDebounce(state.searchTerm, 200)
+    const fetchItems = async (page) => {
```
edoardo, line 92: Perhaps out of scope but there is a way of doing this with `useDataQuery`.
Example here: https://github.com/dhis2/ui/blob/master/components/sharing-dialog/src/autocomplete/sharing-autocomplete.js

edoardo (approved):

PR #889, opened 2021-12-30
fix: repetition infrastructure

janhenrikoverland:
Reducer, action, selectors and tests for `repetition`.


src/reducers/ui.js
```diff
@@ -187,6 +201,7 @@ export const sGetUiActiveModalDialog = (state) =>
     sGetUi(state).activeModalDialog
 export const sGetUiParentGraphMap = (state) => sGetUi(state).parentGraphMap
 export const sGetUiConditions = (state) => sGetUi(state).conditions || {}
+export const sGetUiRepetition = (state) => sGetUi(state).repetition
```
martinkrulltott, line 204:
```suggestion
export const sGetUiRepetition = (state) => sGetUi(state).repetition || {}
```
return empty object to prevent errors being thrown by the consumer if no repetition is found?

martinkrulltott (approved):

src/reducers/ui.js
janhenrikoverland: Should not be needed when EMPTY_UI and DEFAULT_UI has `repetitionByDimension: {}`

PR #891, opened 2022-01-05
refactor: replace AnalyticsResponse with a simplified version

edoardo:
Fixes [TECH-892](https://jira.dhis2.org/browse/TECH-892).

### Key features

1. Implement a simplified processing of the analytics response for LL

---

### Description

The AnalyticsResponse object does a lot of processing of the analytics
response which is not needed for ER Line List.
Instead, only convert booleans values to Yes/No and use ouname column
values for the ou column.

---

### Screenshots

Before:
Screenshot 2022-01-05 at 16 18 51

After:
Screenshot 2022-01-05 at 16 23 59


janhenrikoverland (approved):
LG

PR #892, opened 2022-01-06
feat: numeric conditions to fetch legendsets (TECH-788)

martinkrulltott:

---

### Key features

1. Fetches available legend sets and legends per legend set for numeric conditions

---

### Description

When a numeric dimension is opened, the user has the option to choose a legend set ("`is one of preset options`"). When this is chosen, all available legend sets for that particular dimension is fetched from the api. When in turn a particular legend set is chosen, all legends for that legend set is fetched from the api as well.

Since there are two endpoints for this, one for data elements (e.g. https://debug.dhis2.org/dev/api/dataElements/qrur9Dvnyt5.json?fields=legendSets%5Bid,name%5D) and one for program attributes (e.g. https://debug.dhis2.org/dev/api/trackedEntityAttributes/B6TnnFMgmCk.json?fields=legendSets[id,name]), the query needs to be switched based on the dimension type. The type metadata itself is fetched from the api when opening a saved AO and saved in the `metadata` prop in the redux store.

---

### Known issues

-   [ ] There's no test data for the `PROGRAM_ATTRIBUTE` dimension type, so this part hasn't been tested  

---

### Screenshots

image


src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -73,15 +75,15 @@ const ConditionsManager = ({
     onClose,
     setConditionsByDimension,
 }) => {
-    const dimensionType = dimension.valueType
+    const valueType = dimension.valueType
     const isOptionSetCondition =
-        dimensionType === DIMENSION_TYPE_TEXT && dimension.optionSet
+        valueType === DIMENSION_TYPE_TEXT && dimension.optionSet
 
     const [conditionsList, setConditionsList] = useState(
```
janhenrikoverland, line 82: This piece makes me a bit dizzy..

janhenrikoverland, line 82: This piece makes me a bit dizzy..

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -73,15 +75,15 @@ const ConditionsManager = ({
...
         (conditions.condition?.length &&
             parseConditionsStringToArray(conditions.condition)) ||
             (!conditions.condition?.length &&
-                (SINGLETON_TYPES.includes(dimensionType) ||
+                (SINGLETON_TYPES.includes(valueType) ||
```
janhenrikoverland, line 86: Just wondering why it's called "singleton"

src/components/Dialogs/Conditions/NumericCondition.js
```diff
@@ -249,48 +130,53 @@ const NumericCondition = ({
...
-                        ))}
-                    </SingleSelectField>
-                    {legendSet.legends && (
-                        <MultiSelectField
-                            onChange={({ selected }) =>
-                                setValue(selected.join(';'))
-                            }
-                            inputWidth="330px"
-                            selected={(value?.length && value.split(';')) || []}
+            {operator &&
```
janhenrikoverland, line 133: The first check is not necessary when the second is `operator === OPERATOR_IN`

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -156,16 +176,21 @@ const ConditionsManager = ({
...
             case DIMENSION_TYPE_PERCENTAGE:
             case DIMENSION_TYPE_INTEGER:
             case DIMENSION_TYPE_INTEGER_POSITIVE:
             case DIMENSION_TYPE_INTEGER_NEGATIVE:
             case DIMENSION_TYPE_INTEGER_ZERO_OR_POSITIVE: {
+                if (!availableLegendSets) {
+                    return null
+                }
+
                 const useDecimalSteps =
```
janhenrikoverland, line 191: Maybe avoid the hook-like name.

src/components/Dialogs/Conditions/ConditionsManager.js
janhenrikoverland: I'm not able to read through this multi-line check and understand what it does.

src/components/Dialogs/Conditions/ConditionsManager.js
martinkrulltott: Haha how so? Suggestions welcome 🙂  the logic is that option set conditions are `type === TEXT` + has an `optionSet`

src/components/Dialogs/Conditions/ConditionsManager.js
martinkrulltott: `Singleton` as in _can only be instantiated once_. Some dimensions can have multiple conditions (GT:10;LT:15 etc). They go on separate rows / instances of the corresponding component (e.g. `NumericCondition.js`). However some conditions are single-use, e.g. booleans that will display some checkboxes. You can never display the set of checkboxes multiple times, hence they're "singleton".

src/components/Dialogs/Conditions/ConditionsManager.js
martinkrulltott: Changed to `enableDecimalSteps`

src/components/Dialogs/Conditions/NumericCondition.js
martinkrulltott: Good find, removed 👍

src/components/Dialogs/Conditions/ConditionsManager.js
martinkrulltott: > I'm not able to read through this multi-line check and understand what it does.

No, I agree, it's not super readable. Then again, the logic is quite complex. I could for example extract parts of it separate consts for this purpose only, but I'm not sure if that would make it easier to understand. The key problem is that ConditionsManager handles a lot of different conditions, which themselves vary greatly. Option set, legend set, singleton types and the "regular" types, so it quickly becomes pretty hard to read regardless I think...

src/components/Dialogs/Conditions/ConditionsManager.js
janhenrikoverland: I mean the entire block of logic

janhenrikoverland (approved):
LG, will be thoroughly tested during the next weeks.

PR #894, opened 2022-01-07
fix: input and program redux infrastructure

janhenrikoverland:
Adds reducers, selectors, actions and tests for store.ui props:
- input
- program

For `program` we'll put `name` and `programType` in `store.metadata` by program id.


src/reducers/ui.js
```diff
@@ -85,6 +97,30 @@ export default (state = EMPTY_UI, action) => {
...
+        }
+        case SET_UI_PROGRAM_STAGE: {
+            return {
+                ...state,
+                program: {
+                    ...state.program,
+                    stage: action.value,
+                },
+            }
+        }
```
HendrikThePendric, line 123: In the issue I am currently working on ([TECH-876](https://jira.dhis2.org/browse/TECH-876)), I need to  show a select from which a user can select a program stage (for tracker programs). To populate this select with options, I was thinking of including the `programStages` field when I GET the `programs`. If I go ahead with this we will end up with a `program.stage` and a `program.programStages` property. I would find this a little confusing. Instead we could perhaps only persist the ID and make the name very explicit, i.e. change `stage` (object) to `selectedStageId` (string).

src/reducers/ui.js
```diff
@@ -245,6 +281,11 @@ export const sGetUiRepetition = (state) => sGetUi(state).repetitionByDimension
 
 // Selectors level 2
 
+export const sGetUiInputType = (state) => sGetUiInput(state).type
+
+export const sGetUiProgramId = (state) => sGetUiProgram(state).id
+export const sGetUiProgramStage = (state) => sGetUiProgram(state).stage
```
HendrikThePendric, line 287: Related to my previous comment, if we went with storing the selected stage ID only, we could still have a selector called `sGetUiProgramStage` that gets the stage object, i.e. sth like this:
```js
export const sGetUiProgramStage = (state) =>
    state.ui.program.selectedStageId &&
    state.ui.program.programStages &&
    state.ui.program.programStages.find(
        ({ id }) => id === state.ui.program.selectedStageId
    )
```

HendrikThePendric (commented):
I have a question and a suggestion, see below:

HendrikThePendric (approved):

PR #895, opened 2022-01-10
TECH-876-program-dimensions-accessory-sidebar

HendrikThePendric:
Implements [TECH-876](https://jira.dhis2.org/browse/TECH-876)

---

### Key features

1. Filter dimensions by various criteria:
    - Program
    - Stage (required for tracker programs when input type is "event", optional when input type is "enrollment" and dimension type is "data elements")
    - Search term
    - Dimension type
2. Lock program dropdown

---

### Description

This PR enables users to search for dimension items and renders them to a list. The interactions with the dimension items and their styling is not included in this PR but will be implemented in [TECH-950](https://jira.dhis2.org/browse/TECH-950).

---

### ToDo

- [x] Fix bug which causes app to crash when switching from inputType event to enrolment and the currently selected program is not a tracker program.
- [x] Review used variable names reg. program types and clean up code base.
- [x] Implement paged request with lazy loading
- [x] The new endpoints used to retrieve the program dimension items did not accept a `paging=false` (see [DHIS2-12457](https://jira.dhis2.org/browse/DHIS2-12457)) param and don't support repeating filter syntax (`filter=x:eq:y&filter=a:eq:b`, see [DHIS2-12458](https://jira.dhis2.org/browse/DHIS2-12458)). Once these issues are completed we need to clean up the code in `useProgramDimensions` a bit and adjust the tests.
---

### Known issues

- [ ] It is already possible to lock the program select and clear it. However, currently it is simply locked once a program is selected, which is not correct. The correct behaviour is to lock the program select once one or more program dimension item are selected. This could not be implemented yet because program dimension items selection is not implemented yet (see [TECH-950](https://jira.dhis2.org/browse/TECH-950)).
- [ ] The search term will only match the `name` field, not the `shortName` field. If searching across multiple fields is required, we would need backend support for this.
- [ ] There is some misalignment in the filter UI, which is a result of issue [LIBS-80](https://jira.dhis2.org/browse/LIBS-80) in `@dhis2/ui`. This issue will be addressed very soon and the misalignment will disappear once we upgrade.

---

### Screenshots
Screenshot 2022-01-18 at 10 24 03


src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsFilter.js
```diff
@@ -0,0 +1,76 @@
...
+import { SingleSelect, SingleSelectOption, Input } from '@dhis2/ui'
+import PropTypes from 'prop-types'
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { sGetUiInputType } from '../../../reducers/ui.js'
+import { INPUT_TYPES } from '../InputPanel/index.js'
+import styles from './ProgramDimensionsFilter.module.css'
+import { StageSelect } from './StageSelect.js'
+
+const DIMENSION_TYPES = {
```
martinkrulltott, line 11: Unless `DIMENSION_TYPES` itself is used somewhere I'd expect these consts to be exported individually, e.g. `export const DIMENSION_TYPE_ALL` instead of being props in an object.
Also, super minor, but why the combined export at the bottom instead of just `export const` directly?

martinkrulltott: Same for `PROGRAM_TYPES` in ProgramDimensionPanel.js

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsList.js
```diff
@@ -0,0 +1,103 @@
...
+        const message = searchTerm
+            ? i18n.t("No dimensions found for '{{searchTerm}}'", { searchTerm })
+            : i18n.t('No dimensions found in the {{programName}} program', {
+                  programName,
+              })
+
+        return <div className={styles.noResults}>{message}</div>
+    }
+
+    console.log(isListEndVisible)
```
martinkrulltott, line 72:
```suggestion
```

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
```diff
@@ -0,0 +1,190 @@
...
+    const inputType = useSelector(sGetUiInputType)
+    const selectedProgramId = useSelector(sGetUiProgramId)
+    const selectedStageId = useSelector(sGetUiProgramStage)
+    const setSelectedProgramId = (id) => dispatch(tSetUiProgram(id))
+    const { fetching, error, data, refetch, called } = useDataQuery(query, {
+        lazy: true,
+    })
+    const [searchTerm, setSearchTerm] = useState('')
+    const [dimensionType, setDimensionType] = useState(DIMENSION_TYPES.ALL)
+    const debouncedSearchTerm = useDebounce(searchTerm, 350)
```
martinkrulltott, line 60: `200` is used here https://github.com/dhis2/event-reports-app/blob/26c810f6fc866be4f6d2e83c684471766db0ca3f/src/components/Dialogs/Conditions/OptionSetCondition.js#L100, no idea which is correct, but probably good to be consistent. @cooper-joe ?

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
```diff
@@ -0,0 +1,190 @@
...
+    const filteredPrograms = data?.programs.programs.filter(
+        ({ programType }) =>
+            inputType === INPUT_TYPES.EVENT ||
+            programType === PROGRAM_TYPES.WITHOUT_REGISTRATION
+    )
+    const selectedProgram =
+        filteredPrograms &&
+        selectedProgramId &&
+        filteredPrograms.find(({ id }) => id === selectedProgramId)
```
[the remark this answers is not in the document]
martinkrulltott: and potentially skip `selectedProgramId &&` unless there's a risk for matching `id` with `undefined`?

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
cooper-joe: If the response is not expected to be too "expensive" then a shorter debounce is preferable, but there's flexibility in each use case. Anything between `200` and `500` is acceptable though. If it doesn't really matter resource-wise then let's stick with a consistent `200`.

HendrikThePendric: I think @martinkrulltott makes a good point here about using a consistent debounce time for text inputs across components (and probably the platform as a whole really). But what should this value be?

**TL;DR:** 
I think somewhere between 500-750ms is probably a sensible debounce time for inputs which trigger api requests.

---

**Full motivation**
I find 200ms way too short to debounce an input which triggers an api request. Average time between between key strokes varies between users, contexts and probably characters-location on the keyboard, but the lowest estimate I've found so far mentions 167ms. Our 200ms is way to close to that for this type of use-case. When you do work in the client only, it might make sense to have short debounce time, purely to ensure the UI does not become blocked. But for this type of thing you actually want to choose a debounce time which is a good indication the user has actually finished typing, so you avoid sending too many requests with incomplete search terms.

So, really what we want here is to wait as long as possible, but not so long the user starts thinking the app has become unresponsive. Here's some info about that:
> A delay of less than 100 milliseconds feels instant to a user, but a delay between 100 and 300 milliseconds is perceptible. A delay between 300 and 1,000 milliseconds makes the user feel like a machine is working, but if the delay is above 1,000 milliseconds, your user will likely start to mentally context-switch.
_[Source](https://designingforperformance.com/performance-is-ux/#:~:text=A%20delay%20of%20less%20than,start%20to%20mentally%20context%2Dswitch.)_

So this leads me to the following conclusions:
- The upper bound for a debounce time is 1000ms, if we exceed that the user will become distracted
- The lower bound is a bit more guess work, but I would multiplying the average by 2.5 would be enough to accommodate for differences between users/context/characters, so that `~420` ms.
- So I guess the sweet spot is probably a safe range above the lower and below the upper bound, i.e. *500-750ms*

---
**For reference**
Just to show that choosing a debounce time which is too short can cause really bad issues, please see issue [DHIS2-10628](https://jira.dhis2.org/browse/DHIS2-10628), which was a performance issue (as such, perhaps a bit of an extreme example). The debounce time there was `300` ms to begin with, but because we were getting servers freezing due to too many requests to the users endpoint, we ended up increasing the value to `1200` ms. Now I have to admit that this is a super high value and when using the app, the search does feel a bit sluggish. We have probably taken this a bit too far, but at least it seems to have solved the performance issue....

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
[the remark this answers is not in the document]
HendrikThePendric: > could maybe replace `filteredPrograms &&` with `filteredPrograms?.find`?

I did this

> and potentially skip `selectedProgramId &&` unless there's a risk for matching `id` with `undefined`?

I decided not to do this, since it expresses some intent: this very clearly signals that a selected program is never going to be set when there is no `selectedProgramId`. Omitting it introduces some mental overhead for simple people like me.

Also, it will prevent iterating over the `filteredPrograms` array when this is not needed, but we obviously don't have to start optimising for performance right now, so this is a minor reason.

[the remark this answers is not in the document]
HendrikThePendric: See bdfe9de

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsFilter.js
HendrikThePendric: I did this for `INPUT_TYPES`, `DIMENSION_TYPES` and `PROGRAM_TYPES` in 3124da2

martinkrulltott (approved):

src/components/MainSidebar/ProgramDimensionsPanel/StageSelect.js
```diff
@@ -0,0 +1,39 @@
...
+import { acUpdateUiProgramStage } from '../../../actions/ui.js'
+import { sGetUiProgramStage } from '../../../reducers/ui.js'
+
+const STAGE_ALL = 'STAGE_ALL'
+
+const StageSelect = ({ optional, stages }) => {
+    const dispatch = useDispatch()
+    const selectedStageId = useSelector(sGetUiProgramStage)
+    const onChange = ({ selected }) =>
+        dispatch(acUpdateUiProgramStage(selected))
```
martinkrulltott, line 15: This is a new pattern for us. We usually put the dispatch at the bottom in a `mapDispatchToProps`. Either way is fine I suppose, but its best if we keep a consistent code style. Is this the new preferred way? (so existing components should be refactored?

martinkrulltott: same for `const selectedStageId = useSelector(sGetUiProgramStage)` of course

PR #896, opened 2022-01-10
feat: org unit condition (TECH-883)

martinkrulltott:
Implements [TECH-883](https://jira.dhis2.org/browse/TECH-883)

---

### Key features

1. Implements the org unit condition

---

### Description

Adds support for `valueType: ORGANISATION_UNIT`. Also fixes the scrollbar for the conditions modal so that only the condition content scrolls (everything within the gray box, i.e. not the text description at the top).

---

### Screenshots

image


martinkrulltott:
FYI, to debug this in Chrome dev tools, the ’Show user agent shadow DOM’ setting needs to be enabled https://stackoverflow.com/a/26853319/1581477

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -52,6 +53,7 @@ const DIMENSION_TYPE_TRUE_ONLY = 'TRUE_ONLY'
 const DIMENSION_TYPE_DATE = 'DATE'
 const DIMENSION_TYPE_TIME = 'TIME'
 const DIMENSION_TYPE_DATETIME = 'DATETIME'
+const DIMENSION_TYPE_ORGANISATION_UNIT = 'ORGANISATION_UNIT'
```
janhenrikoverland, line 56: In dhis2 terminology these are "value type", but probably makes sense from an ER pov.

src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -0,0 +1,103 @@
...
+    addMetadata,
+    addParentGraphMap,
+    condition,
+    onChange,
+    rootOrgUnits,
+    metadata,
+    parentGraphMap,
+}) => {
+    const parts = condition.split(':')
+    const value = parts[1]?.length && parts[1]
```
janhenrikoverland, line 26: Pretty simple so maybe not necessary, but ideally a function with a unit test.

janhenrikoverland (approved):
LG

PR #898, opened 2022-01-12
refactor: use vis types from analytics

edoardo:
Related to [TECH-704](https://jira.dhis2.org/browse/TECH-704)

**Requires https://github.com/dhis2/analytics/pull/1126**

---

### Key features

1. import vis types constants and names from analytics
2. move vis type descriptions from analytics to the app
3. enable the vis type filter in `OpenFileDialog`

---

### Description

Because the OpenFileDialog for ER also needs the vis type filter select, the vis types for ER have been defined in `analytics`.
This PR removes the redundant definition in the app and uses the stuff exported from `analytics` wherever possible.

---

### Known issues

-   [x] ~the description for PT shown in the vis type selector might be different from PT in DV~ vis type descriptions have been moved to each app

---

### TODO

- [x] update `analytics` dep once https://github.com/dhis2/analytics/pull/1126 is merged


src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
```diff
@@ -39,12 +45,14 @@ export const VisualizationTypeSelector = ({ visualizationType }) => {
         <Card dataTest={'visualization-type-selector-card'}>
             <div className={classes.listContainer}>
                 <div className={classes.listSection}>
-                    {getVisTypes().map((type) =>
-                        visTypeMap[type].disabled ? (
+                    {[VIS_TYPE_LINE_LIST, VIS_TYPE_PIVOT_TABLE].map((type) =>
```
martinkrulltott, line 48: Consider storing two lists, `availableVisTypes` and `disableVisTypes` (or similar names) somewhere central in the app, so we can easily change it in the future (rather than relying on these harcoded checks everywhere)

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
```diff
@@ -39,12 +45,14 @@ export const VisualizationTypeSelector = ({ visualizationType }) => {
...
+                        type === VIS_TYPE_PIVOT_TABLE ? (
                             <Tooltip
                                 key={`${type}-tooltip`}
                                 placement="bottom"
-                                content={visTypeMap[type].disabledText}
+                                content={i18n.t(
+                                    'Pivot tables are not supported by this app yet'
```
martinkrulltott, line 54: as per the comment above, consider `{{visTypeName}} visualizations are not supported by this app yet` for future-proofing

martinkrulltott (commented):
For scalability and future-proofing, I would consider creating a centralised config for which vis types are supported, so it'll be easier to change when we start supporting more types later.

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
edoardo: Since we know PT is coming, I thought it's ok to have it like this.
I have to refactor anyway when defining the vis types in `modules/visualization`.

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
edoardo: Fixed by using a generic text that does not use the vis type.

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
```diff
@@ -17,20 +22,16 @@ const VisualizationTypeSelector = ({ visualizationType }) => {
...
                 key={type}
                 visType={type}
-                label={visTypeMap[type].name}
-                description={visTypeMap[type].description}
+                label={visTypeDisplayNames[type]}
+                description={visTypeDescriptions[type]}
                 isSelected={type === visualizationType}
-                onClick={isDisabled ? null : handleListItemClick(type)}
-                disabled={isDisabled}
+                onClick={disabled ? null : handleListItemClick(type)}
```
martinkrulltott, line 33:
```suggestion
                onClick={!disabled && handleListItemClick(type)}
```

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
```diff
@@ -89,7 +95,7 @@ const VisualizationTypeSelector = ({ visualizationType }) => {
 }
 
 VisualizationTypeSelector.propTypes = {
-    visualizationType: PropTypes.oneOf(Object.keys(visTypeMap)),
+    visualizationType: PropTypes.oneOf(Object.keys(visTypeDescriptions)),
```
martinkrulltott, line 98: Even though all `visTypes` would have be listed in `visTypeDescriptions`, the source of truth here should be `visTypes`, right? So e.g. 
```suggestion
    visualizationType: PropTypes.oneOf(visTypes.map(visType => visType.type)),
```

martinkrulltott (approved):

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
edoardo: Changed.

src/components/Toolbar/VisualizationTypeSelector/VisualizationTypeSelector.js
edoardo: Changed.

PR #901, opened 2022-01-14
fix: save and fetch options to/from metadata with code (TECH-816)

martinkrulltott:
Implements [TECH-816](https://jira.dhis2.org/browse/TECH-816)

---

### Key features

1. Saves and fetches metadata for options that previously only had `code`

---

### Description

Options use `code` as their identifier instead of `id`. This led to some issue with metadata (`name`) being missing from options that weren't "in scope" in the `Transfer` component (e.g. select an option and then applying a filter which would exclude said option would lead to the name being missing. As seen in the screenshot here: https://github.com/dhis2/event-reports-app/pull/887).

The solution was to make sure that the metadata is added to the store when options were selected plus fetching that metadata for the list of `selected` options.

---

### Known issues

-   [ ] As the `onChange` callback is fired _every_ time a change is made (even if things were only removed), the `addMetadata` action will be called unnecessarily often. However, as the action only appends metadata in a "PUT" maner, there's no impact to the actual store other than from a performance perspective.

### DIFF
--- a/src/api/options.js
+++ b/src/api/options.js
@@ -8,7 +8,7 @@ const optionsQuery = {
         }
 
         return {
-            fields: 'code,name',
+            fields: 'code,name,id',
             filter: filters,
             paging: true,
             page,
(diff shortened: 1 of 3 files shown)


janhenrikoverland (approved):
LG

PR #902, opened 2022-01-17
fix: program indicators as numeric condition (TECH-788)

martinkrulltott:

---

### Key features

1. Display program indicators as numeric condition
2. Refactor names (`DIMENSION_TYPE` -> `VALUE_TYPE`)

---

### Description

Displays program indicators as a numeric condition by consolidating the numeric condition code into a function that's used by `valueType` numeric + `dimensionType` program indicator, both of which can also have legend sets. Also refactors the names of the consts, to differentiate between `valueType` and `dimensionType`, which were mixed up before.


janhenrikoverland (approved):
LG

PR #903, opened 2022-01-19
feat: add hover style to the chip's show more button

HendrikThePendric:
No issue number available

---

### Key features

1. Adds hover style to show-more button in the chip
2. I have removed quite a lot of code that seemed to be redundant
3. Aligns styling of Chip between ER and DV

---

### Description

I mentioned during the test meeting that it isn't very clear for a new user to understand the Chip actually has two distinct click areas, clicking on the "show more icon" does something different than clicking on the main chip area. So I added a hover style to improve that.

Also, I noticed a few other things:
1. `IconButton` has its own style defined in a css module and also accepts a `style` prop. This would lead to unexpected behaviour, for example the hover background color not being applied because the inline non-hover background would take precedence over it.
2. Currently, the `IconButton` is actually only used in one place, so not sure we need it to accept a style prop in the first place. Even if we do need slightly different variants of this in the long run, I think it's better to not mix inline styles with CSS modules too much.
3. The style object passed to the `IconButton` from the `ChipMenu` wasn't actually doing much at all. The color imported from the design system didn't actually exist and the padding was already declared in the css module, so these rules weren't doing anything. The width and height did actually have an effect, so I simply moved them from the style object to the css module.

As a result of the observations above I decided to remove the style prop and related files. I think this is a good move, but it could be that there are new parts of the app I am unaware of, which need a differently styled `IconButton`. If this is the case I might need to reconsider removing the support for inline styles.

---
### Known issues
- [ ] The icon on the left hand side of the Chip still looks off, but this is not due to the chip's styles but the actual SVG. So this still needs to be addressed later on.
---
### Screenshots
Hover style:
Screenshot 2022-01-19 at 14 10 28

Screenshot from events reports app:
Screenshot 2022-01-19 at 14 08 34

Screenshot from data visualizer app:
dv


src/components/IconButton/styles/IconButton.module.css
```diff
@@ -9,8 +9,16 @@
     cursor: pointer;
     padding: 0;
     vertical-align: middle;
+    border-radius: 4px;
```
cooper-joe, line 12: I suggest `border-radius: 2px;` to try to match the outer `border-radius` closer.

src/components/IconButton/styles/IconButton.module.css
```diff
@@ -9,8 +9,16 @@
...
+    height: 22px;
+    margin-left: 2px;
 }
 
 .iconButton:focus {
     outline: none;
 }
+
+.iconButton:hover {
+    background-color: rgba(0, 0, 0, 0.12);
```
cooper-joe, line 23: I suggest two changes:
- `background-color: rgba(77, 182, 172, 0.4);` (`teal300` @ 40%).
- `color: var(--colors-grey900);` to darken the icon on hover.

cooper-joe (changes requested):

cooper-joe (approved):

PR #905, opened 2022-01-19
fix: keep spinner in view when scrolling data-table

HendrikThePendric:
No issue number available

---

### Key features

1. Shows an overlay over the table-body only, as before, so users could theoretically interact with the column headers while the table data is being refreshed.
2. Shows the spinner from a wrapper element which has the same height as the table scollbox, so this will always remain in view

---

### Screenshots

Screenshot 2022-01-19 at 15 31 25

### DIFF
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -11,6 +11,7 @@ import {
     Pagination,
     NoticeBox,
 } from '@dhis2/ui'
+import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React, { useState } from 'react'
 import {
@@ -62,7 +63,7 @@ export const Visualization = ({
                 <NoticeBox error title={i18n.t('Could not load visualization')}>
                     {error?.message ||
                         i18n.t(
-                            'The visualization couldn’t be displayed. Try again or contact your system administrator.'
+                            "The visualization couldn't be displayed. Try again or contact your system administrator."
                         )}
                 </NoticeBox>
             </div>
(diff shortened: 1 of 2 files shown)


src/components/Visualization/Visualization.js
```diff
@@ -80,109 +81,120 @@ export const Visualization = ({
...
-                                            'TEXT',
-                                        {
-                                            digitGroupSeparator:
-                                                visualization.digitGroupSeparator,
-                                            skipRounding: false, // TODO should there be an option for this?
+            <div
+                className={cx(styles.fetchIndicator, {
+                    [styles.fetching]: fetching,
+                })}
+            >
```
HendrikThePendric, line 88: The large diff below is mostly caused because of me adding this wrapper and causing the rest of the code to indent more. There is only one actual change in the diff below, which I will also mark with a comment.

src/components/Visualization/Visualization.js
```diff
@@ -80,109 +81,120 @@ export const Visualization = ({
...
-                                    large={large}
-                                    small={small}
-                                    className={fontSizeClass}
-                                />
-                            )
-                        )}
-                    </DataTableRow>
-                </DataTableHead>
-                {/* https://jira.dhis2.org/browse/LIBS-278 */}
-                <DataTableBody loading={fetching}>
```
HendrikThePendric, line 124: I removed the `loading={fetching}` prop here, because we now have a custom loader UI.

edoardo (approved):

PR #906, opened 2022-01-19
fix: pass headers in the analytics request DHIS2-12314

edoardo:
Relates to [DHIS2-12314](https://jira.dhis2.org/browse/DHIS2-12314)

---

### Key features

1. pass the new `headers` query parameter to analytics request
2. refactor the code for generating the table content

---

### Description

The list of headers to request is directly mapped from the `visualization.columns` and in the same order, so reordering the chips in the layout allows to reorder the data in the rows.

Some columns need special handling during the mapping:
* `ou` is replaced with `ouname`
* `pe` is removed
* all date dimensions selected in the left sidebar should be added

This also simplifies populating the data table and reduces the payload in the analytics response by removing unused headers that otherwise are returned by default.

---

### TODO

-   [ ] once the date selector on the left sidebar is in place, those date fields should be added to `headers`

---

### Screenshots

Showing `ouname` in the same position as `ou`:
Screenshot 2022-01-19 at 15 34 13

### DIFF
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -21,6 +21,22 @@ const formatRowValue = (rowValue, valueType, rowValueItem) => {
     }
 }
 
+const extractHeadersFromColumns = (columns) =>
+    columns.reduce((headers, { dimension }) => {
+        switch (dimension) {
+            case 'pe':
+                break
+            case 'ou':
+                headers.push('ouname')
+                break
+            default:
+                headers.push(dimension)
+                break
+        }
+
+        return headers
+    }, [])
+
 const fetchAnalyticsData = async ({
     analyticsEngine,
     visualization,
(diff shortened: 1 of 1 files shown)


edoardo:
@janhenrikoverland should we add `withHeaders` in `AnalyticsRequestPropertiesMixin`?
Or since it's a ER only thing leave it in the app via the generic `withParameters`?

janhenrikoverland:
@edoardo Since it's not implemented in aggregate analytics yet we can leave it in the app for now, but yeah that could make sense later :+1:

janhenrikoverland:
@edoardo A good old for loop seems to perform much better than reduce still, so even though we use paging I think it makes sense go with for loops here when looping through rows * headers.

edoardo:
> @edoardo A good old for loop seems to perform much better than reduce still, so even though we use paging I think it makes sense go with for loops here when looping through rows * headers.

Refactored here: https://github.com/dhis2/event-reports-app/pull/906/commits/28ad125a7a13d9f2e37f7317cf39e3bdeeb24196

janhenrikoverland (approved):
Approving. Headers camelcase vs lowercase must be handled in the next PR.

PR #908, opened 2022-01-20
fix: alphanumeric case sensitive state (TECH-812)

martinkrulltott:
Implements [TECH-812](https://jira.dhis2.org/browse/TECH-812)

---

### Key features

1. Case sensitive checkbox default state is unchecked
2. Case sensitive checkbox now properly reflects the backend operator (unchecked -> `ILIKE`, checked -> `LIKE`)

image

image


janhenrikoverland (approved):
Seems to work well :+1:

PR #909, opened 2022-01-20
fix: use the correct analytics endpoint based on outputType

edoardo:
**Requires https://github.com/dhis2/analytics/pull/1134**

---

### Key features

1. set `outputType` in current based on the Input selector
2. use the correct analytics endpoint based on `outputType`

---

### Description

Note: in 2.38 only Line List is supported

Line List:
Input: Event -> `/analytics/events/query/...`
Input: Enrollment -> `/analytics/enrollments/query/...`

Pivot Table:
Input: Event -> `/analytics/events/aggregate/...`
Input: Enrollment -> `/analytics/enrollments/aggregate/...`

---

### TODO

-   [x] update analytics dep once https://github.com/dhis2/analytics/pull/1134 is merged

### Screenshots

Example of 2 requests when toggling Input from Event to Enrollment:

Screenshot 2022-01-20 at 16 31 53

### DIFF
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -2,6 +2,7 @@ import { Analytics } from '@dhis2/analytics'
 import { useDataEngine } from '@dhis2/app-runtime'
 import i18n from '@dhis2/d2-i18n'
 import { useEffect, useState, useRef } from 'react'
+import { OUTPUT_TYPE_ENROLLMENT } from '../../modules/visualization.js'
 
 const VALUE_TYPE_BOOLEAN = 'BOOLEAN'
 const VALUE_TYPE_TRUE_ONLY = 'TRUE_ONLY'
@@ -80,7 +81,15 @@ const fetchAnalyticsData = async ({
         }
     }
 
-    const rawResponse = await analyticsEngine.events.getQuery(req)
+    const analyticsApiEndpoint =
+        visualization.outputType === OUTPUT_TYPE_ENROLLMENT
+            ? 'enrollments'
+            : 'events'
+
+    // for 2.38 only /query is used (since only Line List is enabled)
+    const rawResponse = await analyticsEngine[analyticsApiEndpoint].getQuery(
+        req
+    )
 
     return rawResponse
 }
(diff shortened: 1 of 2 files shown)


src/components/Visualization/useAnalyticsData.js
```diff
@@ -80,7 +81,15 @@ const fetchAnalyticsData = async ({
         }
     }
 
-    const rawResponse = await analyticsEngine.events.getQuery(req)
+    const analyticsApiEndpoint =
+        visualization.outputType === OUTPUT_TYPE_ENROLLMENT
```
janhenrikoverland, line 85: There will be more endpoints later, so maybe go with a mapping object right away.

src/modules/current.js
```diff
@@ -20,6 +20,7 @@ export const getDefaultFromUi = (state, action) => {
     return {
         ...state,
         [BASE_FIELD_TYPE]: ui.type,
+        outputType: ui.input.type,
```
janhenrikoverland, line 23: This requires knowledge about exact variable names in the redux store outside of the reducer file. Could we use a selector?

src/components/Visualization/useAnalyticsData.js
janhenrikoverland: Maybe analyics endpoint could be added to the `outputTypeMap` in https://github.com/dhis2/event-reports-app/blob/develop/src/modules/visualization.js#L29?

src/modules/current.js
edoardo: I was wondering exactly the same and checked other parts of the code, no selectors are used.
But we do have a selector for `ui.input.type` and probably all the other too.

src/components/Visualization/useAnalyticsData.js
edoardo: Ok, I guess having this mapping in `useAnalyticsData` is fine since I don't expect it to be needed in other places.

src/modules/current.js
janhenrikoverland: Right, I really think we should use selectors all over.

src/components/Visualization/useAnalyticsData.js
janhenrikoverland: Ok

janhenrikoverland (approved):
Approved when updated.

src/modules/current.js
edoardo: It's not possible to use selectors because they need the whole `state` of the store.
Here only the `current` section of the state is available.

janhenrikoverland (approved):

PR #910, opened 2022-01-20
feat: your dimensions accessory sidebar

HendrikThePendric:
Implements [TECH-877](https://jira.dhis2.org/browse/TECH-877)

---

### Key features

1. Your dimensions panel with search input
2. Extracted `DimensionsList` component which is now shared between the program and your dimension panel
3. Partially implemented list items and their interactions

---

### Description

This PR add the "Your dimensions" panel with a search input and adds some basic interactions to the dimension list items. The basic notion is to merge this PR into develop ASAP so other team member can use these dimension panels. The dimension items will be refined in a subsequent PR.

---

### Known issues

- Dimension list items don't have all their states (i.e. locked / disabled / drag-only) implemented yet (deferred to another PR). 
- Dimension list items don't show the correct icons for all dimension types yet. (not yet available at the moment).
- Selection count badge has not been implemented yet

---

### Screenshots

Screenshot 2022-01-24 at 16 03 51


src/actions/ui.js
```diff
@@ -100,9 +100,10 @@ export const tClearUi = () => (dispatch, getState) => {
     )
 }
 
-export const acSetUiActiveModalDialog = (value) => ({
+export const acSetUiActiveModalDialog = (value, metadata) => ({
```
martinkrulltott, line 103: Unsure about this pattern of having `metadata` as an additional prop in the request here, especially since its not entirely clear what the metadata is referring to (renaming the action to `acSetUiActiveDimensionModal` might help with that). Is it preferred to use a consolidated action like this or should the consumer just do a separate "add metadata" call instead? @janhenrikoverland

src/components/MainSidebar/DimensionsList/DimensionsList.js
```diff
@@ -0,0 +1,111 @@
...
+import styles from './DimensionsList.module.css'
+
+const getNoResultsMessage = (searchTerm, programName) => {
+    if (searchTerm) {
+        return i18n.t("No dimensions found for '{{searchTerm}}'", {
+            searchTerm,
+        })
+    }
+    if (programName) {
+        i18n.t('No dimensions found in the {{programName}} program', {
```
martinkrulltott, line 15:
```suggestion
        return i18n.t('No dimensions found in the {{programName}} program', {
```

src/components/MainSidebar/DimensionsList/DimensionsList.js
```diff
@@ -0,0 +1,111 @@
...
+
+    if (loading) {
+        return (
+            <div className={styles.loaderWrap}>
+                <CircularLoader small />
+            </div>
+        )
+    }
+
+    if (error) {
```
martinkrulltott, line 53: technically `if` might be correct here, but when reading through a component with multiple `return`s, I think it make sense to wrap them in `if`, `if else` and finally `else` to indicate that they're indeed separate paths.

martinkrulltott: same for the returns in `getNoResultsMessage` above

martinkrulltott: IMO superior to all of this is to avoid multiple returns altogether, assign the value to a `let` in an `if / if else / else` block and use a single return at the end though.

src/actions/ui.js
HendrikThePendric: > or should the consumer just do a separate "add metadata" call instead?

If this is the preferred way, we should probably remove the redux middleware.

martinkrulltott: No I like the redux middleware, just want it to be clear what the metadata is referring to. So best would be to rename the action to `acSetUiActiveDimensionModal` so its obvious that its a "dimension" that is open (hence said dimension can have metadata related to it, which is passed as an optional prop). Also, whats a "ModalDialog" anyway? 😄

src/components/MainSidebar/DimensionsList/DimensionsList.js
HendrikThePendric: This comment threw me a bit. In react components I see `if/return` it much more than `if/else`. I'm happy to comply with the code conventions of the team, but then I think the team is choosing a pattern that's a bit different from the react ecosystem as a whole....

src/actions/ui.js
martinkrulltott: one could also argue that "active" is too loose. "open" makes more sense (as thats what modals do, open and close, not activate and deactivate)... so `acSetUiOpenDimensionModal` 🎉

martinkrulltott, line 103:
```suggestion
export const acSetUiOpenDimensionModal = (value, metadata) => ({
```

src/components/MainSidebar/DimensionsList/DimensionsList.js
```diff
@@ -0,0 +1,111 @@
...
+                    {error?.message ||
+                        i18n.t(
+                            'There was a problem loading the dimensions. Try again, or contact your system administrator.'
+                        )}
+                </NoticeBox>
+            </div>
+        )
+    }
+
+    if (dimensions.length === 0) {
```
martinkrulltott, line 66:
```suggestion
    if (!dimensions?.length) {
```
`dimensions` isn't `isRequired` and has no `defaultProps` value (speaking of which, this in itself is a lint warning I think?), could thus be `undefined`?

src/components/MainSidebar/DimensionsList/DimensionsList.js
jenniferarnesen: I don't think we ever agreed on a team convention. I prefer the `if/return` as well, and if that is more compliant with the react ecosystem, then it makes sense to me to stick with it.

src/components/MainSidebar/DimensionsList/DimensionsList.js
```diff
@@ -0,0 +1,111 @@
...
+    loading,
+    fetching,
+    error,
+    dimensions,
+    programName,
+    searchTerm,
+    setIsListEndVisible,
+}) => {
+    const scrollBoxRef = useRef()
+    const updateIsListEndVisible = () => {
```
martinkrulltott, line 32: the name `updateIsListEndVisible` sort of clashes with `setIsListEndVisible` (but looking at it further I understand that the former is a wrapper for the latter. Was thinking it might be called `onScroll`, but that might not fit so well when called from the `useEffect` below, or does it? 🤔

src/components/MainSidebar/DimensionsList/DimensionsList.js
HendrikThePendric: Once loading is done, and there is no error, then we have a dimensions array. I think rendering a "no results found message" when there is a dimension array which is empty, expresses the intent of what this block of code does very well.

The suggested change makes the code more robust, but I would prefer it to fail fast.

Hope you are OK with keeping this as-is.

src/components/MainSidebar/DimensionsList/DimensionsList.js
janhenrikoverland: I understand both arguments (different paths vs react component convention).

I often think `if / else if` gets messy when the clauses are unrelated.

janhenrikoverland: When checking for different values of the same varible then `if / else if` is pretty readable, like `switch / case`.

janhenrikoverland: Lets stick to the react component convention in this case.

src/components/MainSidebar/DimensionsList/DimensionsList.js
HendrikThePendric: I see what you mean. Tried to avoid naming collisions in 6ccaedc but not sure that's much better....

martinkrulltott: haha, try reading `setIsListEndVisible(isScrollboxEndVisible(scrollBoxRef.current))` quickly 5 times in a row 😉  Hard to say, I think this way at least you can't mix up the two fns, but neither is a clear choice maybe. Up to you 👍

src/components/MainSidebar/DimensionsList/DimensionsList.js
martinkrulltott: Ok, if you want it to fail if it (for whatever reason) doesn't have a `dimension` at all then maybe that warrants a `try catch` so we can provide a user-friendly error? 
I mean, the argument "it doesn't fail now so we don't need it to be more robust" only holds up for so long, right, so the undefined-check was just a suggestion to provide some future-proofing (+ personally I think checking `!array.length` is just as expressive as `array.length === 0` so I think your intent is still very clear).

src/components/MainSidebar/DimensionsList/DimensionsList.js
martinkrulltott: I don't mind the `if/return` if it's a single use (e.g. commonly to check for a prop and return `null` to avoid multiple partial renders while loading). But if it's simply the case of checking for different conditions and returning primitive values (i18n still counts in this case) like it is with `getNoResultsMessage` I don't see why not to use `if / else if / else`.

src/components/MainSidebar/DimensionsList/DimensionListItem.js
```diff
@@ -0,0 +1,107 @@
...
+                [styles.recommended]: recommended,
+                [styles.dragOnly]: dragOnly,
+            })}
+            onClick={onClick}
+        >
+            <div className={styles.icon}>
+                <Icon />
+            </div>
+            <div className={styles.label}>{name}</div>
+            {locked ? (
```
martinkrulltott, line 85: why ternary with `''` instead of just `&&`?

src/components/MainSidebar/ProgramDimensionsPanel/useProgramDimensions.js
```diff
@@ -55,6 +61,11 @@ const reducer = (state, action) => {
                 error: action.payload,
                 dimensions: null,
             }
+        case ACTIONS.SET_LIST_END_VISIBLE:
```
martinkrulltott, line 64: Same as we discussed the other day, is `ACTIONS` ever used in a standalone fashion? Or could these props just be extract to individual consts e.g. `const ACTION_RESET = 'RESET'`?

src/components/MainSidebar/DimensionsList/DimensionListItem.js
HendrikThePendric: Oops.... Some copy-pasted code from [the design specs](https://github.com/dhis2/design-specs/blob/master/src/components/dimension-list.js)... Well spotted. Updated in 9a65b339b9314778eb71d167bb4913ab587ff6f1

src/components/MainSidebar/YourDimensionsPanel/YourDimensionsPanel.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types'
+import React, { useState } from 'react'
+import { useDebounce } from '../../../modules/utils.js'
+import { DimensionsList } from '../DimensionsList/index.js'
+import { useYourDimensions } from './useYourDimensions.js'
+import styles from './YourDimensionsPanel.module.css'
+
+const YourDimensionsPanel = ({ visible }) => {
+    const [searchTerm, setSearchTerm] = useState('')
+    const debouncedSearchTerm = useDebounce(searchTerm, 500)
```
martinkrulltott, line 12: Now that `useDebounce` is used in multiple places, maybe the default debounce time should be exported as a separate prop? (e.g. `useDebounce(searchTerm, DEFAULT_DEBOUNCE_TIME)`) or let `useDebounce` set the default time internally and only pass `searchTerm` without an explicit time (e.g. `useDebounce(searchTerm)`)?

HendrikThePendric: 🤔 I think we might need to export two values, one for when a user is typing, and the other for when we want to respond to events that fire often (i.e. `mousemove`, `resize`, etc). Maybe like this:
```js
export const DEFAULT_USER_INPUT_DELAY = 500
export const DEFAULT_WINDOW_EVENT_DELAY = 150
```
FYI: I already used the shorter time yesterday for setting the table max-width after a window resize.

src/components/Visualization/useAnalyticsData.js
```diff
@@ -136,6 +139,7 @@ const useAnalyticsData = ({
     page,
     pageSize,
 }) => {
+    const uiState = useSelector(sGetUi)
```
martinkrulltott, line 142: can we avoid passing the whole `ui` object around and only pick out the parts that are actually used? (looks like `program` and `input`)

martinkrulltott: also, you already have `visualization` (which is `current`, right?), wouldn't that include the parts needed anyway?

martinkrulltott: image
debugging this I see that `program` and `input` are missing from `visualization`, but that's just because they haven't been added yet, right? so once they're added to `visualization` you could just access everything that `uiState` provides through that prop?

src/components/Visualization/useAnalyticsData.js
```diff
@@ -168,6 +173,7 @@ const useAnalyticsData = ({
                 mounted.current && setData({ headers, rows, pageCount, total })
                 onResponseReceived(analyticsResponse)
             } catch (error) {
+                console.error(error)
```
martinkrulltott, line 176: just left-over from debugging or is this our preferred way of logging errors?

src/components/MainSidebar/YourDimensionsPanel/YourDimensionsPanel.js
martinkrulltott: Sounds good to me 👍

martinkrulltott: I just saw one being `150` and figured we could have a `default` and a custom. But two "defaults" works as well

jenniferarnesen:
@HendrikThePendric Should the i18n/en.pot file be in the PR?

src/components/Visualization/useAnalyticsData.js
HendrikThePendric: This was an excellent point. The fact that these values were not present in the first place was the problem, so reading them directly from the UI state was not really a solution. In 7d50f1b I have ensured that the required properties get added to `current` from `ui`.

src/components/MainSidebar/YourDimensionsPanel/YourDimensionsPanel.js
HendrikThePendric: > or let `useDebounce` set the default delay internally and only pass `searchTerm` without an explicit time (e.g. `useDebounce(searchTerm)`)?

Actually this was a very nice idea. I've gone with that.

HendrikThePendric:
> @HendrikThePendric Should the i18n/en.pot file be in the PR?

@jenniferarnesen There is actually 1 new string in there. The others have been moved around because I extracted a component and moved it to another folder. But yeah, we do need it in for that one string.

src/modules/current.js
```diff
@@ -21,11 +21,19 @@ export const getDefaultFromUi = (current, action) => {
...
         outputType: ui.input.type,
+        ...getProgramFromUi(ui),
+        ...getProgramStageFromUi(ui),
         ...getAxesFromUi(ui),
         ...getOptionsFromUi(ui),
     }
 }
 
+export const getProgramFromUi = (ui) =>
+    ui.program?.id ? { program: { id: ui.program.id } } : null
```
martinkrulltott, line 32:
```suggestion
    ui.program?.id && { program: { id: ui.program.id } }
```

martinkrulltott: same for `getProgramStageFromUi` below

martinkrulltott (approved):

jenniferarnesen (approved):

PR #911, opened 2022-01-21
fix: improved loading state and response handling for legend sets

martinkrulltott:
Fix for [TECH-788](https://jira.dhis2.org/browse/TECH-788)

---

### Key features

Prevents elements from "popping up" and improves the transparency of the loading state by:
1. Storing and fetching legend set responses to/from the `metadata` store
2. Using the `loading` prop in the `select`s to indicate the loading state
3. Setting the `placeholder` to `Loading...` to indicate the loading state

---

### Description

Instead of fetching `legendSets` and `legends` every time, the response from the first request is saved in the store (under `metadata`) and is used when the modal is reopened.
To indicate the loading state, both the `loading` prop in the `select`s themselves are used as well as the `placeholder` which now says `Loading...`. The latter prevents selected options from popping up once the loading is complete (without it the field would look like there were no selected options while the fetching was being made).

---

### Screenshots

_loading state_
image

_placeholder_
image


janhenrikoverland (approved):
`ui` issues to be addressed separately.

PR #912, opened 2022-01-24
fix: optionSets display value

edoardo:
### Key features

1. show optionSets item name instead of code

---

### Description

The analytics response returns the optionSets item code instead of the id.
This means the lookup on `metaData.items` by id is not possible in this case, but the name must be looked up from the code instead.
This is not optimal as the search has to loop through `metaData.items` until one matches.

### Screenshots

Before:
Screenshot 2022-01-24 at 14 42 59

After:
Screenshot 2022-01-24 at 14 42 03


martinkrulltott (approved):
LGTM. Would be good to get @janhenrikoverland 's approval too though

janhenrikoverland (approved):
LG

PR #913, opened 2022-01-25
feat: forward compatibility: pe - time dimension conversion TECH-796

edoardo:
Implements [TECH-869](https://jira.dhis2.org/browse/TECH-869) and [TECH-960](https://jira.dhis2.org/browse/TECH-960)

---

### Key features

1. convert `pe` dimension to a time dimension when opening a "old format" AO [(TECH-869)](https://jira.dhis2.org/browse/TECH-869)
2. format column values to pass to `headers` in the analytics request ([TECH-960](https://jira.dhis2.org/browse/TECH-960))
3. pass the time dimensions as arguments in the analytics request
4. set default sorting to first column ascending

---

### Description

The new time dimensions that can be added to Columns/Filters require special handling in the analytics request.
1. they need to be passed in `headers` like the other dimensions but lowercased
2. they need to be passed as query arguments in the analytics request instead of as `dimension` argument

---

### TODO

-   [x] use constants for the time dimension names
-   [ ] refactor some code? move `isTimeDimension` and `getAdaptedVisualization` to `modules/visualization`?

---

### Known issues

-   [x] backend is not ready with the support for the new time dimensions so the analytics request resolves in a conflict due to missing start/end dates
-   [ ] Layout is empty in LL due to this code returning `false` because `layout` has the replaced period dimension but there isn't a correspondent entry in `metaData` (ie. `enrollmentDate` in layout and `pe` in metaData)

### DIFF
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -5,6 +5,11 @@ import { useEffect, useState, useRef } from 'react'
 import {
     OUTPUT_TYPE_ENROLLMENT,
     OUTPUT_TYPE_EVENT,
+    TIME_DIMENSION_EVENT_DATE,
+    TIME_DIMENSION_ENROLLMENT_DATE,
+    TIME_DIMENSION_INCIDENT_DATE,
+    TIME_DIMENSION_SCHEDULED_DATE,
+    TIME_DIMENSION_LAST_UPDATED,
     headersMap,
 } from '../../modules/visualization.js'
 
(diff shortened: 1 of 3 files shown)


src/components/Visualization/useAnalyticsData.js
```diff
@@ -35,22 +43,66 @@ const formatRowValue = (rowValue, header, metaDataItems) => {
...
+
+const getAdaptedVisualization = (visualization) => {
+    const adaptedColumns = []
+    const adaptedRows = []
+    const adaptedFilters = []
+    const headers = []
+    const timeDimensionParameters = {}
+
+    visualization.columns.forEach((dimensionObj) => {
+        const dimensionId = dimensionObj.dimension
```
martinkrulltott, line 63: since this is repeated thrice with the only difference being which array its pushing to in the end, maybe its worth extracting this to a generic function that's called for each axis with an input array to know where to push the end result? (e.g. `adaptedColumns.push(dimensionObj)` is replaced by `outputArray.push(dimensionObj)`)

src/components/Visualization/useAnalyticsData.js
```diff
@@ -61,10 +113,15 @@ const fetchAnalyticsData = async ({
     sortField,
     sortDirection,
 }) => {
+    // XXX must be reviewed when PT comes around. Most likely LL and PT have quite different handling
```
martinkrulltott, line 116:
```suggestion
    // TODO: must be reviewed when PT comes around. Most likely LL and PT have quite different handling
```

martinkrulltott (approved):

src/components/Visualization/useAnalyticsData.js
edoardo: Refactored.

src/components/Visualization/useAnalyticsData.js
edoardo: Fixed.

PR #915, opened 2022-01-25
feat: expand and collapse layout panel

jenniferarnesen:
Implements [TECH-957](https://jira.dhis2.org/browse/TECH-957)

---

### Key features

1. Expand/collapse layout panel
2. Tiny little refactor snuck in: changed prop "location" to "initialLocation" since it is more accurate and won't be confused with arguments to functions named "location" in the App component

---

### Description

Add a button that toggles the layout between expanded and collapsed state. Both expanded and collapsed will grow and shrink according to content in the Layout, but in collapsed state, the max it can grow to is 20vh. In expanded state, it will grow to fill 100% of the parent container if needed. Once that max has been reached, then a scrollbar is shown. Here's an example:

1. click on New:  height 78px, expanded state
2. add dimensions: height grows to a max of 100% of parent container height, then shows scrollbar
3. click Update: height 20vh (or less - down to 78px - depending on content), collapsed state
4. click expand: height set to show all content up to max of 100% parent container height
5. click collapse: height 20vh (or less - down to 78px - depending on content), collapsed state

* when a new ER: start in expanded state (assumption is that user will be adding dimensions and configuring the ER)
* when existing ER is opened: start in collapsed state

---

### Screenshots

Open an existing ER
https://user-images.githubusercontent.com/6113918/150996275-c8510462-f512-4073-a3de-785c6d0531ab.mov

Existing is open and make changes and update:
https://user-images.githubusercontent.com/6113918/150995968-e3f9be26-8c39-4fd3-83c0-131ca66ba139.mov

Create a new ER:
https://user-images.githubusercontent.com/6113918/150995976-6597cf1c-64df-49e9-8290-5abd4b74b468.mov


cooper-joe (approved):
LGTM! 🎉 

Note: I think it might be a good idea to disable the toolbar in some cases, but I'm not sure which yet. Potentially when in expanded but not more than `20vh`. However, that can easily be added in a new PR once I've taken a look at it in more depth, so I'm happy to approve this.

src/components/Layout/styles/Layout.module.css
```diff
@@ -0,0 +1,41 @@
...
+    flex-direction: column;
+}
+
+.overflowContainer {
+    overflow-x: hidden;
+    overflow-y: auto;
+    flex: 1;
+    height: auto;
+    min-height: 78px;
+    max-height: 20vh;
```
HendrikThePendric, line 12: I think instead of a `min-height` you could declare a `flex-basis` of `78px` and then set `flex-shrink` to 0, to prevent the item to shrink beyond its base size.
I also think that `height` is redundant here. So probably I'd change this part to the following (using the [`flex` shorthand property](https://developer.mozilla.org/en-US/docs/Web/CSS/flex)):
```suggestion
    flex-basis: 1 0 78px;
    max-height: 20vh;
```

HendrikThePendric (approved):
Looks great. I left two nit-picky suggestions about how to do conditional classes with the `classnames` package and CSS modules, and a potentially more flexboxy set of CSS rules..

PR #916, opened 2022-01-25
TECH-874-main-sidebar-dimension-selection

HendrikThePendric:
Implements [DHIS2-874](https://jira.dhis2.org/browse/DHIS2-874)

---

### Key features

1. Hooks up various main and accessory sidebar panels to the global store
2. Updates the UI state from a loaded visualization
3. Updates the current state based on user selections in the sidebar
4. Addresses various issues which were discovered during testing

---

### Description

This PR enables dimension item selection in the main sidebar. 

---

### TODO

- [x] Ensure the selected program and its stages stages are put into the metadata store, both when selecting things in the sidebar, but also when loading a saved visualization.
- [x] Implement the selected state of `DimensionListItem` and show selected items.
- [x] Populate `DimensionListItem` secondary text with programs which use the same data-element across different stages.
- [x] Implement badges with selection count for "Program dimensions" and "Your dimensions"
- [x] Set sidebar selection from saved visualisation, when loading one
- [x] When input is enrollment, we should only display tracker programs (`programType: 'WITH_REGISTRATION`), this seems to be the wrong way around at the moment.
- [x] Switching input should always trigger a "clear" (this should already work, but needs to be verified) / Tweak logic for clearing program dimension filter inputs, see [Slack thread](https://dhis2.slack.com/archives/G01PW9YGZD2/p1643281376001400)
- [x] Page size should be 50 for lazy loading lists
- [x] The stage field should be indented slightly, see specs
- [x] ~Select and Button should have equal height (check with @joecooper first)~ Just add a `small` prop to the button for now
- [x] Implement time dimension selection logic 
- [x] Read input from AO when one is loaded
- [x] Prepare for all dimension items having `stage.id`

---

### Screenshots
Screenshot 2022-02-02 at 16 46 53


src/modules/dimensionTypes.js
```diff
@@ -0,0 +1,56 @@
...
+])
+
+/**TIME**/
+export const DIMENSION_TYPE_EVENT_DATE = 'eventDate'
+export const DIMENSION_TYPE_ENROLLMENT_DATE = 'enrollmentDate'
+export const DIMENSION_TYPE_INCIDENT_DATE = 'incidentDate'
+export const DIMENSION_TYPE_SCHEDULED_DATE = 'scheduledDate'
+export const DIMENSION_TYPE_LAST_UPDATED = 'lastUpdated'
+
+export const DIMENSION_TYPES_TIME = new Set([
```
jenniferarnesen, line 50: These are also defined in `modules/timeDimensions`. We should consolidate somehow.

src/reducers/ui.js
```diff
@@ -105,6 +106,15 @@ export default (state = EMPTY_UI, action) => {
                 program: EMPTY_UI.program,
             }
         }
+        case CLEAR_UI_STAGE: {
+            return {
+                ...state,
+                program: {
+                    id: state.program.id,
+                    stage: undefined,
```
jenniferarnesen, line 114: This property is now named `stageId` and accordingly action UPDATE_UI_PROGRAM_STAGE_ID. So this action should also be renamed (and clear the value of `stageId`)

src/components/MainSidebar/MainSidebar.js
```diff
@@ -19,6 +25,43 @@ const TAB_INPUT = 'INPUT'
 const TAB_PROGRAM = 'PROGRAM'
 const TAB_YOUR = 'YOUR'
 
+const useSelectedDimensions = () => {
```
jenniferarnesen, line 28: What about exporting this from a separate hook file and letting the components use the hook, rather than passing the `isSelected` function around?

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsList.js
```diff
@@ -24,19 +24,20 @@ const ProgramDimensionsList = ({
...
             searchTerm={searchTerm}
         />
     )
 }
 
 ProgramDimensionsList.propTypes = {
     inputType: PropTypes.string.isRequired,
-    programId: PropTypes.string.isRequired,
-    programName: PropTypes.string.isRequired,
+    isSelected: PropTypes.func.isRequired,
```
jenniferarnesen, line 39: Was confused, thought this was a boolean, which didn't make sense. Maybe a name that suggests it is a function (I know it's not as pretty, but `getIsSelected` or similar)

src/components/MainSidebar/__tests__/__snapshots__/TimeDimensions.spec.js.snap
```diff
@@ -26,23 +26,11 @@ exports[`input type ENROLLMENT, WITH_REGISTRATION, custom labels 1`] = `
...
-      </div>
-      <div
-        class="menuTrigger"
-      >
-        <svg
-          height="16"
-          viewBox="0 0 16 16"
-          width="16"
-          xmlns="http://www.w3.org/2000/svg"
+        <span
```
jenniferarnesen, line 29: The change to the snapshots is the removal of the `menuTrigger` button. Just confirming that is correct.

src/reducers/ui.js
jenniferarnesen: I'm wondering if the properties of the default `program` property ought to include the `id` and `stageId` (set to undefined) to make clear the expected structure.

src/actions/ui.js
```diff
@@ -19,6 +19,7 @@ import {
     UPDATE_UI_PROGRAM_ID,
     UPDATE_UI_PROGRAM_STAGE_ID,
     CLEAR_UI_PROGRAM,
```
edoardo, line 21: Are these 2 actually clearing the ids? Perhaps rename them to `CLEAR_UI_*_ID` as well.

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
```diff
@@ -101,17 +105,22 @@ const ProgramDimensionsPanel = ({ visible }) => {
...
             programType === PROGRAM_TYPE_WITHOUT_REGISTRATION
         ) {
-            const artificialStageId = selectedProgram.programStages[0].id
-            const metadata = {
-                [artificialStageId]: selectedProgram.programStages[0],
-            }
-            dispatch(acUpdateUiProgramStageId(artificialStageId, metadata))
+            const artificialStage = selectedProgram.programStages[0]
+            const metadata = { [artificialStage.id]: artificialStage }
+            dispatch(acUpdateUiProgramStage(artificialStage.id, metadata))
```
edoardo, line 123: This has been renamed
```suggestion
            dispatch(acUpdateUiProgramStageId(artificialStage.id, metadata))
```

src/modules/input.js
```diff
@@ -0,0 +1,3 @@
+export const getInputFromVisualization = (vis) => ({
```
edoardo, line 1: Perhaps rename to:
```suggestion
export const getInputTypeFromVisualization = (vis) => ({
```

src/components/MainSidebar/MainSidebar.js
HendrikThePendric: I agree that passing the isSelected function around isn't ideal. On the other hand, if we let individual components use this hook, it means that several "instances" of this hook will be doing the exact same computation, which also isn't great. I briefly considered to implement this using a context-provider + hook. This is probably a good solution, because it would allow us to do the computations in one place, and allow all children to import the hook and thus avoid passing things around. I originally decided against it because it would make the code slightly harder to follow. But, upon reflection, it seems that passing props through several layers of components doesn't make for the most readable code ever. So I'll try a solution with a context provider.

HendrikThePendric: See implementation in [19bc100](https://github.com/dhis2/event-reports-app/pull/916/commits/19bc100b45b16f6cb89c21325adbebdb4f5a6795)

jenniferarnesen (approved):
Lgtm

src/components/MainSidebar/MainSidebar.js
jenniferarnesen: I like the implementation.

edoardo (approved):

PR #917, opened 2022-01-27
fix: various bugs with conditions (TECH-787)

martinkrulltott:
Implements [TECH-787](https://jira.dhis2.org/browse/TECH-787)

---

### Key features

1. Adds metadata to prevent optionSets from crashing when creating a new AO + prevent `addMetadata` to be fired for already added options
2. Throws a proper error for unsupported dimension types (fix to prevent this from actually happening TBD)
3. Filters out unsupported valueTypes
4. Adds the valueType and dimensionType to the modal title for test purposes

### DIFF
--- a/src/components/Dialogs/Conditions/OptionSetCondition.js
+++ b/src/components/Dialogs/Conditions/OptionSetCondition.js
@@ -78,13 +78,18 @@ const OptionSetCondition = ({
     const dataTest = 'option-set'
 
     const setValues = (selected) => {
-        addMetadata(
-            state.options
-                .filter((item) => selected.includes(item.code))
-                .map((item) => ({
+        state.options
+            .filter(
+                (item) =>
+                    selected.includes(item.code) &&
+                    !selectedOptions.find((so) => so.code === item.code)
+            )
+            .forEach((item) => {
+                addMetadata({
                     [item.id]: item,
-                }))
-        )
+                })
+            })
+
         onChange(`${OPERATOR_IN}:${selected.join(';') || ''}`)
     }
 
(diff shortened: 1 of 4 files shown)


src/components/Dialogs/Conditions/OptionSetCondition.js
```diff
@@ -78,13 +78,18 @@ const OptionSetCondition = ({
...
     const setValues = (selected) => {
-        addMetadata(
-            state.options
-                .filter((item) => selected.includes(item.code))
-                .map((item) => ({
+        state.options
+            .filter(
+                (item) =>
+                    selected.includes(item.code) &&
+                    !selectedOptions.find((so) => so.code === item.code)
```
martinkrulltott, line 85: As `Transfer` returns all selected options, even when only a subset has been added (e.g. if `1` is already selected and `2` is selected by clicking, then both `1` and `2` will be returned as selected), this prevents `addMetadata` to be fired for "old" selected options (`1` in the example above)

Before: `1` is preselected, `2` is now selected -> `addMetadata(1), addMetadata(2)`
Now: `1` is preselected, `2` is now selected -> `addMetadata(2)`

src/components/Dialogs/Conditions/OptionSetCondition.js
```diff
@@ -78,13 +78,18 @@ const OptionSetCondition = ({
...
+            )
+            .forEach((item) => {
```
janhenrikoverland, line 87: Build object and dispatch once.

martinkrulltott: fixed with `reduce`

janhenrikoverland (approved):

PR #923, opened 2022-02-01
fix: set stage to default object in case metadata is not present

jenniferarnesen:
Implements [TECH-804](https://jira.dhis2.org/browse/TECH-804)
---

### Key features

1. fix crash due to missing object. If metadata is missing then `stage` should be set to a default object

### DIFF
--- a/src/components/MainSidebar/TimeDimensions.js
+++ b/src/components/MainSidebar/TimeDimensions.js
@@ -35,6 +35,7 @@ const TimeDimensions = () => {
     const stageId = useSelector(sGetUiProgramStage)
     const program =
         useSelector((state) => sGetMetadataById(state, programId)) || {}
+    const stage = useSelector((state) => sGetMetadataById(state, stageId)) || {}
     const timeDimensions = getTimeDimensions()
 
     const dimensionIds = [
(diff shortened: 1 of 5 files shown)


src/components/MainSidebar/TimeDimensions.js
```diff
@@ -22,6 +22,9 @@ import { PROGRAM_TYPE_WITH_REGISTRATION } from './ProgramDimensionsPanel/Program
 import styles from './TimeDimensions.module.css'
 
 const getName = (dimension, program, stage) => {
+    if (!dimension.nameParentProperty) {
```
edoardo, line 25: This makes it much easier to follow, although probably it would work anyway if the metaData is fixed.

src/components/MainSidebar/TimeDimensions.js
```diff
@@ -35,20 +38,13 @@ const TimeDimensions = () => {
     const stageId = useSelector(sGetUiProgramStage)
```
edoardo, line 38: Should we rename this here?
```suggestion
    const stageId = useSelector(sGetUiProgramStageId)
```

edoardo (approved):

src/components/MainSidebar/TimeDimensions.js
jenniferarnesen: Yes, I just figures since `LastUpdated` doesn't have a custom name property, that it is more correct to do it this way, even though the result is the same.

src/components/MainSidebar/TimeDimensions.js
jenniferarnesen: Yes I just figured I'd do that in a separate PR (which I'll do right now)

PR #925, opened 2022-02-01
feat: repeatable events (TECH-894)

martinkrulltott:
Implements [TECH-894](https://jira.dhis2.org/browse/TECH-894)

**Requires https://github.com/dhis2/analytics/pull/1152**

---

### Key features

1. Adds repeatable events to the conditions modal

---

### Description

Displays the repeatable tab in the conditions modal when a data element is opened and input type is enrollment. The tab will be disabled if the data element's stage isn't repeatable.
The input values are stored in `ui` and are subsequently parsed to the backend format when setting `current` from `ui` (previously created by @janhenrikoverland ).

The analytics request takes the backend format for repetition and transforms this into separate dimensions on the request, (Analytics PR https://github.com/dhis2/analytics/pull/1152)

**One-liner bonus features**

1. Removes the `valueType` check for option sets
2. Removes the `programStageId` from the `legendSet` query

---

### Known issues

-   [ ] Saving / loading repetition hasn't been implemented or tested yet, as there seems to be a backend bug for this at the moment

---

### Screenshots

`DATA_ELEMENT` + `ENROLLMENT` + `stage.repeatable === true`

image

`DATA_ELEMENT` + `ENROLLMENT` + `stage.repeatable === false`

image

`!DATA_ELEMENT` or `!ENROLLMENT`

image


HendrikThePendric (approved):

src/api/legendSets.js
```diff
@@ -74,11 +74,14 @@ export const apiFetchLegendSetsByDimension = async ({
         default:
             throw new Error(`${dimensionType} is not a valid dimension type`)
     }
+
+    const [id] = dimensionId.split('.').reverse()
```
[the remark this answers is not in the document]
martinkrulltott: This was actually based on @edoardo 's way of reading the `programStageId` and the `dimensionId` from a string that can be either "dimensionId" or "programStageId.dimensionId", but since I don't need the `programStageId` I just omitted that part of the destruction. Maybe not the most readable, but it's short! 🙂

src/components/Dialogs/Conditions/ConditionsManager.js
```diff
@@ -461,6 +531,15 @@ const mapStateToProps = (state, ownProps) => ({
         sGetUiConditionsByDimension(state, ownProps.dimension?.id) || {},
     dimensionIdsInLayout: sGetDimensionIdsFromLayout(state),
     displayNameProp: sGetSettingsDisplayNameProperty(state),
+    inputType: sGetUiInputType(state),
+    stage:
+        sGetMetadataById(
+            state,
+            ownProps.dimension?.id?.substring(
```
edoardo, line 538: Could it be that `dimension.id` does not contain the program stage id?
In that case this code won't work.

edoardo (approved):

src/components/Dialogs/Conditions/ConditionsManager.js
martinkrulltott: In the case the dimension doesn't have a stage prefix (i.e. there is no stage) then `{}` is returned and there will be no `stage` for this dimension...

PR #927, opened 2022-02-02
fix: remove state in AppWrapper that was causing multiple instances of redux store

jenniferarnesen:
### Key features

Buggy behaviour when loading an ER. Sometimes layout would be populated, and sometimes not. The issue was that there are two separate redux stores being read/written.

---

### Description

Fix was to remove state (ouLevels). ouLevels is not currently in use anyway, but a comment was also added to remind us to not add state to this component in the future.

### DIFF
--- a/src/AppWrapper.js
+++ b/src/AppWrapper.js
@@ -41,6 +41,13 @@ const providerDataTransformation = (rawData) => {
     }
 }
 
+/*
+ * The redux store is being created here and this should only happen once,
+ * because having multiple store instances leads to very unpredictable behaviour.
+ * To avoid having multiple stores, ensure this component only renders once,
+ * so it should remain stateless and be the app's most outer container.
+ */
+
 const AppWrapper = () => {
     const engine = useDataEngine()
     const store = configureStore([
--- a/src/AppWrapper.js
+++ b/src/AppWrapper.js
@@ -1,10 +1,6 @@
-import {
-    apiFetchOrganisationUnitLevels,
-    CachedDataQueryProvider,
-} from '@dhis2/analytics'
+import { CachedDataQueryProvider } from '@dhis2/analytics'
 import { useDataEngine } from '@dhis2/app-runtime'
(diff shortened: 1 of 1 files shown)


src/AppWrapper.js
```diff
@@ -45,46 +45,52 @@ const providerDataTransformation = (rawData) => {
...
     }, [])
 
+    return (
+        <CachedDataQueryProvider
+            query={query}
+            dataTransformation={providerDataTransformation}
+        >
+            <App
+                initialLocation={history.location}
+                ouLevels={ouLevels} // TODO: Unused by App.js?
```
edoardo, line 70: This is not used at the moment in ER.
I think it's safe to remove the whole part, but it can be done in a separate PR if preferred.

src/AppWrapper.js
```diff
@@ -45,46 +45,52 @@ const providerDataTransformation = (rawData) => {
...
+}
+
+AppWrapperInner.propTypes = {
+    engine: PropTypes.object,
+}
+
+const AppWrapper = () => {
+    const engine = useDataEngine()
+    const store = configureStore([
+        thunk.withExtraArgument(engine), // FIXME: Not needed for ER? Pointed out by @edoardo
```
edoardo, line 83: I don't think I pointed out this one.
Although it should not be needed once all fetch functions in `api` are not used anymore.

edoardo (approved):

HendrikThePendric (approved):
I think this solution is probably best for now. However, it was hard to get to the root cause of this problem and I don't really want to have to experience it all again. So, we should add a comment above line 36 to avoid this type of problem being reintroduced. Maybe something like this:
```
/*
 * The redux store is being created here, and this should only happen once,
 * because having multiple store instances leads to very unpredictable behaviour.
 * To avoid having multiple stores, ensure this component only renders once,
 * so it should remain stateless and be the app's most outer container.
 */
 ```

PR #934, opened 2022-02-07
feat: error handling (TECH-706)

martinkrulltott:
Implements [TECH-706](https://jira.dhis2.org/browse/TECH-706)

---

### Key features

1. No data error
2. Layout validation (no period error, no columns error,  no org unit error, no program error)
3. Replace manual "Update" buttons with the `UpdateVisualizationContainer`
4. Refactor all errors from classes to consts

---

### Description

Adds layout validation and error handling for when the backend returns no data. All "Update" buttons are replaced by the `UpdateVisualizationContainer` that handles both the validation, load errors and other basic tasks automatically.

---

### TODO

-   [ ] The history section of the `UpdateVisualizationContainer` was commented out for now, should this be used in ER?

---

### Known issues

-   [ ] period validation is disabled as these dimensions can't be added to the layout properly yet

---

### Screenshots

_no data_

image

_no org unit_

image

_no columns_

image

_no program_

image


src/modules/error.js
```diff
@@ -53,6 +57,46 @@ export class VisualizationNotFoundError extends VisualizationError {
...
+        )
+    }
+}
+
+export class NoColumnsError extends VisualizationError {
+    constructor() {
+        super(
+            EmptyBox,
+            i18n.t('Columns is empty'),
+            i18n.t('At at least one item to Columns.')
```
edoardo, line 75: Typo?
```suggestion
            i18n.t('Add at least one item to Columns.')
```

src/components/UpdateButton/UpdateVisualizationContainer.js
```diff
@@ -0,0 +1,68 @@
...
+    acSetLoadError,
+    acClearLoadError,
+    acSetVisualizationLoading,
+} from '../../actions/loader.js'
+import { acSetShowExpandedLayoutPanel } from '../../actions/ui.js'
+import { GenericClientError } from '../../modules/error.js'
+import { validateLayout } from '../../modules/layoutValidation.js'
+import { sGetCurrent, sGetCurrentFromUi } from '../../reducers/current.js'
+
+// TODO: Replace all 'Update' buttons throughout the app with UVC
```
edoardo, line 13: Is there any Update button not replaced left?

src/components/VisualizationOptions/VisualizationOptionsManager.js
```diff
@@ -37,25 +25,25 @@ const VisualizationOptionsManager = ({ onUpdate }) => {
...
+VisualizationOptionsManager.propTypes = {}
 
 const mapStateToProps = () => ({})
 
-const mapDispatchToProps = {
-    onUpdate: tSetCurrentFromUi,
-}
+const mapDispatchToProps = {}
 
 export default connect(
```
edoardo, line 48: Does it need to be connected at all?

src/modules/layoutValidation.js
```diff
@@ -0,0 +1,61 @@
...
+    // }
+
+    if (!layout?.program?.id) {
+        throw new NoProgramError()
+    }
+}
+
+export const validateLayout = (layout) => {
+    switch (layout.type) {
+        case VIS_TYPE_LINE:
```
edoardo, line 57:
```suggestion
        case VIS_TYPE_LINE_LIST:
```

edoardo (commented):
I added some comments and there are a bunch of linting issues.

src/modules/error.js
martinkrulltott: Yepp, good find!

src/components/UpdateButton/UpdateVisualizationContainer.js
martinkrulltott: Nope, comment removed 👍

src/components/VisualizationOptions/VisualizationOptionsManager.js
martinkrulltott: True, forgot to remove these. Fixed 👍

src/modules/error.js
```diff
@@ -53,6 +57,46 @@ export class VisualizationNotFoundError extends VisualizationError {
...
+export class NoPeriodError extends VisualizationError {
+    constructor() {
+        super(
+            PeriodError,
+            i18n.t('No time dimension selected'),
+            i18n.t('Add at least one time dimension to the layout.')
+        )
+    }
+}
+
```
HendrikThePendric, line 99: Could be there are reasons for going with this class based approach I'm not aware of, but I find this rather confusing. Because these are classes with an `Error` suffix, we see them invoked as `new BlaBlaError()`. This firstly makes me think that these classes are [custom error classes](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/error#custom_error_types) which we could throw, and secondly makes me suspect the returned object comes with methods we might want to call.

It seems to me that all we want to have is a bunch of simple functions that produce simple objects with properties `icon`, `title`, and `description`. If this is in fact all we need, I would be in favour of abandoning this class based approach in favour of some simple functions that return object.

HendrikThePendric (commented):
I'll leave it up to @edoardo to approve/request changes, but I also reviewed and have a few reservations about how the errors were implemented. Left some code comments for that.

src/modules/layoutValidation.js
```diff
@@ -0,0 +1,70 @@
...
+    if (!isAxisValid(axis)) {
+        throw error
+    }
+}
+
+const validateLineListLayout = (layout) => {
+    validateAxis(layout.columns, new NoColumnsError(layout.type))
+    validateDimension(
+        layoutGetDimension(layout, DIMENSION_ID_ORGUNIT),
+        new NoOrgUnitError(layout.type),
```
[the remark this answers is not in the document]
martinkrulltott: Good find, copy-paste error from DV, where the message is dependent on the vis type

src/modules/error.js
martinkrulltott: I think the class based approach (originally in DV https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/modules/error.js) was to unify the errors, both from an input (`visType`) and output (params) perspective. But as we don't use any input in ER (as all errors use static text) the only thing it really does is to make sure what to expect back from the object (`title`, `icon`, `description`). Im not sure if class is the best way to do this though, but having some sort of typed interface would be convenient I suppose so we have a clear pattern to follow. Suggestions are welcome

edoardo: We adopted many things from DV, but we don't have to keep them if there are simpler and better solutions.
I always thought ER as a opportunity to review stuff we did in DV and improve and eventually bring those improvements back to DV.

HendrikThePendric: I'm not sure what you mean about "typing", but I'll just give a very brief demo of what I was trying to suggest:
```js
// CURRENT
export class VisualizationError {
    constructor(icon, title, description) {
        this.icon = icon
        this.title = title
        this.description = description
    }
}

export class GenericClientError extends VisualizationError {
    constructor() {
        super(
            GenericError,
            genericErrorTitle,
            i18n.t('There is a problem with this visualization.')
        )
    }
}
// SUGGESTION
export const visualizationError (icon, title, description) => ({
    icon,
    title,
    description,
})

export const genericClientError () => visualizationError({
    GenericError,
    genericErrorTitle,
    i18n.t('There is a problem with this visualization.')
})
```

martinkrulltott: Fixed with https://github.com/dhis2/event-reports-app/pull/934/commits/db6096acf1c240e3b0a56494f4633d33cf6f1881

HendrikThePendric (approved):

PR #936, opened 2022-02-08
feat(visualization): improve usage of available space and loading state

HendrikThePendric:
Implements [DHIS2-TECH-961](https://jira.dhis2.org/browse/DHIS2-TECH-961)

---

### Key features

1. Implement sticky pagination which is always in view and never exceeds the scroll container's width
2. Implement auto-height, so that the data-table occupies the full vertical available space
3. Improve loading behaviour

---

### Description

There's not much more to say about the sticky navigation and auto-height features, but the loading behaviour improvements require more explanation. The feature I wanted to implement was to clear the data and thus the data-table when a new visualization is loaded. I had to solve a few bugs along the way:
- The `App` component wasn't populating its `previousLocation` with the current location on mount, which was causing some checks in the `history.listen` callback to behave differently than expected
- The `needsReresh` function seemed unnecessary: on first load the fetch needs to happen anyway and since each subsequent refetch was triggered via the `history.listen` callback which includes a superset of the checks found in `needsRefetch`. As such I decided to remove it completely, because it was preventing the refetch to happen when reloading the current visualization
- Finally I realised that the state that seemed to be attached to the location in `MenuBar` wasn't actually doing anything. Perhaps this code was still using the old `history` API, but in any case it was not possible to read any state from the location
- After fixing these bugs, I managed to implement the loading behaviour I wanted. To also trigger a refetch when reloading the current visualization, I had to add a history listener to `useAnalyticsData`

### Screenshots

Screenshot 2022-02-08 at 17 05 06


src/components/Visualization/useAnalyticsData.js
```diff
@@ -200,51 +203,65 @@ const useAnalyticsData = ({
...
+            const pageCount = analyticsResponse.metaData.pager.pageCount
+            const total = analyticsResponse.metaData.pager.total
+
+            mounted.current && setError(undefined)
+            mounted.current && setData({ headers, rows, pageCount, total })
+            onResponseReceived(analyticsResponse)
+        } catch (error) {
+            mounted.current && setError(error)
+        } finally {
+            // console.log('finally?', mounted.current)
```
edoardo, line 236: Remove this comment:
```suggestion
```

edoardo (approved):
Quite a lot going on.
Code looks good, we'll have to test thoroughly all the navigation scenarios.

PR #937, opened 2022-02-09
feat: dnd layout and dimensions panel

jenniferarnesen:
Implements [TECH-957](https://jira.dhis2.org/browse/TECH-957)
---

### Key features

1. Implement drag and drop in Layout and between dimensions panel and layout

---

### Description

1. Implement dragging from the Main and Time dimensions panels into an axis in the layout, into the desired position
2. Improved dnd within the layout (reordering and moving between axes). It is now much easier to place an item exactly where you want it.

For the implementation we switched from react-beautiful-dnd to dnd-kit, which supports "grid" dnd. Currently only dragging of a single item is possible. Dragging to remove an item from the layout has not been implemented here.

#### Overview of some of the files
1. DndContext. This component must be a parent of any draggable and droppable components. It handles the `onDragEnd` event, which is where redux dispatches are made to update the layout and the dimensions. It also is where the DragOverlay component is, which is what's shown to be dragging.
2.  DraggableDimensionItem. This is the same as DimensionItem with the addition of draggable functionality. Eventually DimensionItem will be removed since all dimension items are supposed to be draggable (still present until Your and Program dimensions have been converted to draggable)

---

### TODO in upcoming PRs
1. implement dnd in Your dimensions and Program dimensions
2. improve on the drop behaviour (tweak the collision strategy)
3. Look for performance improvements (sometimes dragging can be laggy)

### Screenshots

https://user-images.githubusercontent.com/6113918/153815342-f6ffedfe-28c6-4ad1-aea4-91d64ff4c96f.mov


janhenrikoverland (approved):
LG, let's merge and test.

PR #938, opened 2022-02-09
chore: metadata for time and main dimensions

jenniferarnesen:
Implements additional work for:
- [TECH-804](https://jira.dhis2.org/browse/TECH-804)
- [TECH-962](https://jira.dhis2.org/browse/TECH-962)

---

### Key features

1. Add (static) main and time dimensions to the metadata store when the app loads
2. Implemented selector hooks to access dynamic main and time dimensions. The dynamic part of these are the `disabled` property and for time dimensions the `name` as well.
3. Let sidebar components use these hooks

----

### Update

After discussing a little bit with @martinkrulltott it became clear that we actually need to manually maintain the data in the metadata store, as opposed to just dumping stuff in there. The features of this PR can now be described as follows:
1. Old program and stages are cleared from the metadata store when input or program change
2. Time dimension names are updated when input, program, or stage change
3. (Program) stages are available directly in the metadata store

I initially agreed with @martinkrulltott that I would also "sanitise" the metadata objects before adding them to the metadata store, see [this Slack thread](https://dhis2.slack.com/archives/G0108CD1421/p1644504757889399?thread_ts=1644423176.125979&cid=G0108CD1421) for motivation. However, while implementing this part I noticed it was making things unnecessarily complex. If we start removing properties from objects before we add them to the metadata store, we would need to distinguish between "this program" (i.e. `state.metadata[programId]`) and "that program" (i.e. a `propgram` originating from programs-dropdown), where the former had fewer properties than the latter. I discussed with @janhenrikoverland that I would skip that part, so apart from also making nested objects (i.e. program stages) available directly in the metadata, I am not doing any data transformations. Everything is pretty much stored as it is returned to us from the server.


HendrikThePendric:
There are 2 points in this PR I think we should discuss:
1. We're still adding the "static" main and period dimensions to the metadata store, but if you were to read these dimensions directly from there, the `disabled` field would not be present, and for time dimensions the `name` might be incorrect. When using the hooks you are guaranteed to always get the correct dimension "state". So I'm considering perhaps not adding them to the metadata store at all (the hooks don't need them in the metadata store).
2. Thinking back of the meeting this morning, I seem to remember that @martinkrulltott would like a single way to access any dimension. This PR does not meet that requirement. But I could write an additional hook that would make that possible. Let me know if that would be convenient/needed?

src/components/MainSidebar/__tests__/__snapshots__/TimeDimensions.spec.js.snap
```diff
@@ -1645,7 +1645,7 @@ exports[`stage missing 1`] = `
           <span
             class="primary"
           >
-            le enrollment date
+            Tracking date
```
jenniferarnesen, line 1648: Something changed here - it is not using the custom name.

src/reducers/ui.js
```diff
@@ -321,3 +331,69 @@ export const sGetUiConditionsByDimension = (state, dimension) =>
...
+                ),
+                disabled: !enabledDimensionIds.has(dimension.id),
+            }
+            return acc
+        }, {})
+    }, [programId, inputType, stageId])
+}
+
+export const useTimeDimension = (id) => {
+    const timeDimensions = useTimeDimension()
```
jenniferarnesen, line 397: This looks like a mistake? Should it be `useTimeDimensions` ?

src/reducers/ui.js
```diff
@@ -321,3 +331,69 @@ export const sGetUiConditionsByDimension = (state, dimension) =>
...
+        const enabledDimensionIds = getEnabledTimeDimensionIds(
+            inputType,
+            program,
+            stage
+        )
+
+        return Object.values(timeDimensions).reduce((acc, dimension) => {
+            acc[dimension.id] = {
+                ...dimension,
+                name: getTimeDimensionName(
```
edoardo, line 384: I'm not quite sure about this part.
Is `metadata` synchronised to reflect the same name when it changes?

src/components/MainSidebar/MainDimensions.js
```diff
@@ -1,78 +1,28 @@
...
-                selected={getIsDimensionSelected(
-                    DIMENSION_TYPE_LAST_UPDATED_BY
-                )}
-            />
+            {Object.values(mainDimensions).map(
+                ({ id, name, dimensionType, disabled }) => (
+                    <DimensionItem
+                        key={id}
+                        dimensionType={dimensionType}
+                        name={name}
```
edoardo, line 19: For main dimensions here the names are not coming from metadata, but I guess in this case they are the same anyway...

src/modules/metadata.js
```diff
@@ -21,12 +23,14 @@ const getFixedDimensions = () => ({
 
 export default function () {
```
jenniferarnesen, line 24: Could we switch this to a named export?

src/components/MainSidebar/__tests__/__snapshots__/TimeDimensions.spec.js.snap
HendrikThePendric: Well spotted. This code is outdated now, but there was indeed an issue here.

jenniferarnesen:
Looks like something might be getting entered wrong in the metadata. This screenshot shows properties directly in the metadata redux property:
image

src/modules/timeDimensions.js
```diff
@@ -1,48 +1,105 @@
...
 import {
     DIMENSION_TYPE_EVENT_DATE,
     DIMENSION_TYPE_ENROLLMENT_DATE,
     DIMENSION_TYPE_INCIDENT_DATE,
     DIMENSION_TYPE_SCHEDULED_DATE,
     DIMENSION_TYPE_LAST_UPDATED,
 } from './dimensionTypes.js'
+import { OUTPUT_TYPE_EVENT } from './visualization.js'
 
 export const NAME_PARENT_PROPERTY_PROGRAM = 'program'
```
jenniferarnesen, line 13: I don't think this needs to be exported. Seems to only be used in this file.

src/modules/dimensionTypes.js
```diff
@@ -47,6 +47,7 @@ export const DIMENSION_TYPE_INCIDENT_DATE = 'incidentDate'
 export const DIMENSION_TYPE_SCHEDULED_DATE = 'scheduledDate'
 export const DIMENSION_TYPE_LAST_UPDATED = 'lastUpdated'
 
+// TODO - not used?
```
jenniferarnesen, line 50: Should we just remove this const since it isn't used?

jenniferarnesen:
LGTM.

HendrikThePendric (approved):

PR #939, opened 2022-02-09
feat: add line list icon on start screen

janhenrikoverland:
Implements [TECH-970](https://jira.dhis2.org/browse/TECH-970)

Using the line list icon in "Your most viewed event reports".


src/components/Visualization/StartScreen.js
```diff
@@ -11,26 +14,66 @@ import { sGetLoadError } from '../../reducers/loader.js'
...
+}
+
+const StartScreen = ({ error, username, setLoadError }) => {
+    const getContent = () => {
+        const data = useMostViewedVisualizations(username)
+
+        if (data.error) {
+            setLoadError(data.error)
+            return null
+        }
```
HendrikThePendric, line 69: I think this is fine, but you could also consider handling this from the `onError` callbacks of the `useDataQuery` hook options. I think it'd be effectively the same, but might look a bit more conventional than first calling an action and then rendering content.

HendrikThePendric (approved):
LGTM 👍 I left one minor suggestion in the code. 

Also I wanted to point out that in #934 I suggested to change the errors from class instances to simple functions. If that PR gets merged some things will break, such as:
```js
if (error instanceof VisualizationError) {
    // do sth
}
```

PR #945, opened 2022-02-13
feat: program and event status modal (TECH-697)

martinkrulltott:
Implements [TECH-697](https://jira.dhis2.org/browse/TECH-697)

---

### Key features

1. Add support for program status and event status in the dimension modal

---

### Description

Adds support for program status and event status when opening a dimension and the dimension modal is displayed. 
Based on this [mockup](https://www.figma.com/proto/dSCBFauVvYqrdhvQE4tBxM/Event-Reports?page-id=201%3A10293&node-id=132%3A7680&show-proto-sidebar=1&starting-point-node-id=132%3A7680).
~The values are stored in the redux store as props for `ui`: `programStatus` and `eventStatus`.~
The selected items are stored normally in the redux store under `ui.itemsByDimension`.

---

### TODO

-   [ ] ~connect `ui` <-> `current` for `ui.eventStatus` and `ui.programStatus`~ not needed anymore since we now store item ids normally in `ui.itemsByDimension`
- [ ] Screenshots don't match 100% the design sketches

---

### Screenshots

image

image


src/components/Dialogs/FixedDimension.js
```diff
@@ -3,13 +3,28 @@ import {
...
 import { connect } from 'react-redux'
 import { tSetCurrentFromUi } from '../../actions/current.js'
 import { acAddMetadata } from '../../actions/metadata.js'
 import { acSetUiItems, acAddParentGraphMap } from '../../actions/ui.js'
+import {
+    DIMENSION_TYPE_EVENT_STATUS,
+    DIMENSION_TYPE_PROGRAM_STATUS,
+} from '../../modules/dimensionTypes.js'
 import { removeLastPathSegment, getOuPath } from '../../modules/orgUnit.js'
+import {
```
jenniferarnesen, line 19: Wondering if this the best module for these constants or if another location would be more appropriate. If it is only used by FixedDimension, perhaps it can just be internal to that component?

src/reducers/ui.js
```diff
@@ -76,6 +78,8 @@ export const DEFAULT_UI = {
     },
     itemsByDimension: {
         [DIMENSION_ID_ORGUNIT]: [],
+        [DIMENSION_TYPE_EVENT_STATUS]: [], // XXX should this default be set in the component instead? since these are fixed dimensions perhaps this default here makes sense
```
jenniferarnesen, line 81: remember to remove comment. I personally like having defaults in redux so the components don't have to worry about it. But not too opinionated on that.

jenniferarnesen (approved):

src/components/Dialogs/FixedDimension.js
```diff
@@ -70,8 +88,97 @@ const FixedDimension = ({
...
+                            dense
+                            className={classes.verticalCheckbox}
+                        />
+                    ))}
+                </div>
+            </>
+        )
+    }
+
+    const renderEventStatus = () => {
```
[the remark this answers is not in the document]
martinkrulltott: Yeah I considered that as well, but I figured since we'll never have more than these two and since the amount of common lines are quite low compared to the unique stuff I thought it would just over-complicate things (even though there's clearly a common pattern here)

src/reducers/ui.js
edoardo: We can leave it there then and remove the comment.
```suggestion
        [DIMENSION_TYPE_EVENT_STATUS]: [],
```

src/components/Dialogs/FixedDimension.js
edoardo: One constant is used in `modules/visualization`.
I'm not against moving them, they are used for both program and event stage.

edoardo (approved):

PR #947, opened 2022-02-15
chore: fix prgram stage filter

HendrikThePendric:
Implements issue nr not available

---

### Key features

1. Use `id:startsWith:<ID>` instead of `id:like:<ID>` to filter data element dimensions
2. Ensure the stage ID is not added to the global store when `STAGE_ALL` is selected

---

### Known issues

-   [ ] We are sending the correct query params to the server but the filter still doesn't seem to be working.

---


src/components/MainSidebar/ProgramDimensionsPanel/useProgramDimensions.js
```diff
@@ -99,7 +99,7 @@ const createDimensionsQuery = ({
     ) {
         // This works because data element IDs have the following notation:
```
edoardo, line 100: I'm not sure about this comment.
The format we should store the dimension (at least in `current`) needs to be:
`{
dimension: [dimensionId],
programStage: { id: [programStageId] }
}`

edoardo (approved):

PR #952, opened 2022-02-17
fix: set periodType field on period dimensions to `PERIOD`

HendrikThePendric:
Quick fix so all period dimensions have period type `PERIOD`

### DIFF
--- a/src/modules/visualization.js
+++ b/src/modules/visualization.js
@@ -4,6 +4,7 @@ import {
     AXIS_ID_FILTERS,
     VIS_TYPE_LINE_LIST,
     VIS_TYPE_PIVOT_TABLE,
+    DIMENSION_ID_PERIOD,
 } from '@dhis2/analytics'
 import i18n from '@dhis2/d2-i18n'
 import { DEFAULT_CURRENT } from '../reducers/current.js'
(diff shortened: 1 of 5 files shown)


src/modules/visualization.js
```diff
@@ -92,22 +86,13 @@ export const transformVisualization = (visualization) => {
...
 
-const transformDimensions = (dimensions, { outputType, type }) =>
+const transformDimensions = (dimensions) =>
     dimensions.map((dimensionObj) => {
         if (dimensionObj.dimensionType === 'PROGRAM_DATA_ELEMENT') {
             return {
                 ...dimensionObj,
                 dimensionType: DIMENSION_TYPE_DATA_ELEMENT,
             }
-        } else if (
```
edoardo, line 102: We still need this part, see https://jira.dhis2.org/browse/TECH-869.

edoardo (approved):

src/modules/visualization.js
```diff
@@ -92,22 +86,13 @@ export const transformVisualization = (visualization) => {
...
-            dimensionObj.dimension === DIMENSION_ID_PERIOD &&
-            type === VIS_TYPE_LINE_LIST
-        ) {
-            return {
-                ...dimensionObj,
-                dimension: outputTypeTimeDimensionMap[outputType],
-                dimensionType: DIMENSION_ID_PERIOD,
```
jenniferarnesen, line 109: But this line (109) should probably be removed, right?

edoardo: Apparently we do need `dimensionType: PERIOD` when saving, currently I get an error from the backend due to `pe` being passed as `dimensionType`.

edoardo (approved):

PR #967, opened 2022-02-21
feat: show dimension context menu when hovering item in dimension panel

jenniferarnesen:
### Key features

1. Hovering over a dimension item, to the right of the label, reveals the dimension menu. Clicking on it opens the same menu as that which is on the layout chips.

---

### Description

File notes:
1. DimensionMenu/DimensionMenu was renamed to DimensionMenu/MenuItems. Same code - no changes.
2. ChipMenu was moved/renamed to DimensionMenu/DimensionMenu since it is generic and can be used both for layout chips and dimension items. Functionality was added to determine the axis from store ui layout.
3. Instead of passing visType all over the place, get the value in the DimensionMenu itself, where the value is needed.

This PR does not address issues with fonts/margins/padding. That will be a separate PR.

---

### Screenshots

https://user-images.githubusercontent.com/6113918/155079115-b4474299-0897-4772-916b-e581634c3949.mov


janhenrikoverland (approved):
Let's merge and test.

PR #974, opened 2022-02-22
fix: lookup in metadata for column header names

edoardo:
### Key features

1. lookup column header names in metadata

---

### Description

Time dimensions are changing their name depending on the program selection.
This should reflect in the data table as well, in the column header, the name should match the one in the layout and in the side panel.

### Screenshots

Before:
Screenshot 2022-02-22 at 17 01 16

After:
Screenshot 2022-02-22 at 16 58 38


janhenrikoverland (approved):
LG

janhenrikoverland (approved):

PR #983, opened 2022-02-24
fix: time dimension improvements

HendrikThePendric:
Implements:
- [TECH-979](https://jira.dhis2.org/browse/TECH-979)
- A number of KFMT issues related to period/time dimensions

---

### Key features

1. Use names from metadata to set initially selected period items ([TECH-979](https://jira.dhis2.org/browse/TECH-979))
2. Tweak default time dimension selection, see [Slack thread](https://dhis2.slack.com/archives/G3TL0J145/p1645628057636699)
3. Clear time dimensions from layout when switching Input / program, see [Slack thread](https://dhis2.slack.com/archives/D9N90FGCQ/p1645613457823179)

---

### Description

Below some details about the key features:
1. Since the backend is now including the names of the period dimension items in the `metadata.items` in the analytics response, it is now possible to start using these names for the intitially selected period dimensions.
2. The changes to the default time selection caused one test to break because in one of the tests it was enabling one extra time dimension. I have adjusted the expected values in the test, because I believe the following is correct, could you please confirm @janhenrikoverland?
    ```js
    {
        // inputType, program and stage args for getEnabledTimeDimensionIds
        inputType: OUTPUT_TYPE_ENROLLMENT,
        program: {
            programType: PROGRAM_TYPE_WITH_REGISTRATION,
            displayIncidentDate: true,
        },
        stage: {
            id: '1',
            hideDueDate: false,
        },
        // expected enabled dimensions
        expected: [
            DIMENSION_TYPE_EVENT_DATE, // NEW ADDITION IN THIS PR
            DIMENSION_TYPE_ENROLLMENT_DATE,
            DIMENSION_TYPE_INCIDENT_DATE,
            DIMENSION_TYPE_LAST_UPDATED,
        ],
    },
    ```
3. At first we were simply clearing all the time dimensions when the input or program changed or was cleared. However, I discussed with @janhenrikoverland that it would also be possible to keep some of the time dimensions in place. This seemed like a good idea so that's what I've implemented now: the time dimensions that are going to get disabled are getting removed and the other ones are kept in place. We couldn't think of any situations where this would be problematic, but could have overlooked something, so please pay extra attention to that part when reviewing.

---

### Screenshots

*Names from metadata*
Screenshot 2022-02-24 at 09 56 27
Screenshot 2022-02-24 at 09 56 15

*Default selection*
Screenshot 2022-02-24 at 09 54 53

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -1,13 +1,11 @@
-import {
-    DIMENSION_TYPES_PROGRAM,
-    DIMENSION_TYPES_TIME,
-} from '../modules/dimensionTypes.js'
+import { DIMENSION_TYPES_PROGRAM } from '../modules/dimensionTypes.js'
 import {
     getDefaulTimeDimensionsMetadata,
     getDynamicTimeDimensionsMetadata,
     getProgramAsMetadata,
 } from '../modules/metadata.js'
 import { PROGRAM_TYPE_WITH_REGISTRATION } from '../modules/programTypes.js'
+import { getEnabledTimeDimensionIds } from '../modules/timeDimensions.js'
 import { OUTPUT_TYPE_EVENT } from '../modules/visualization.js'
 import { sGetMetadataById } from '../reducers/metadata.js'
 import { sGetRootOrgUnits } from '../reducers/settings.js'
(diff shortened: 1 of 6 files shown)


edoardo (approved):

janhenrikoverland:
@HendrikThePendric A comment on 2) in the description:

The combination of enrollment + tracker program should not enable "event date": https://jira.dhis2.org/secure/attachment/23247/23247_Screenshot+from+2021-11-26+11-53-11.png

HendrikThePendric:
> @HendrikThePendric A comment on 2) in the description. Event date should not be enabled for enrollment: https://jira.dhis2.org/secure/attachment/23247/23247_Screenshot+from+2021-11-26+11-53-11.png

Well spotted @janhenrikoverland, it turned out I made a silly copy-paste error which broke this test in the first place. Corrected in https://github.com/dhis2/event-reports-app/pull/983/commits/e67cbbdf14332e7883fd72bb908c453f5089243c

PR #986, opened 2022-02-24
fix: hide your dimensions button when no dimensions are available

edoardo:
### Key features

1. hide your dimensions button if no your dimensions are available

---

### Description

Moved the menu item for Your dimensions into a separate component which queries the `dimensions` endpoint to check if any dimension is available.
If not, hide the button as clicking it would result in an empty list in the panel.

### DIFF
--- a/src/components/MainSidebar/YourDimensions.js
+++ b/src/components/MainSidebar/YourDimensions.js
@@ -5,18 +5,18 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { MenuItem } from './MenuItem/index.js'
 import {
-    yourDimensionsResource,
-    yourDimensionsFilter,
+    YOUR_DIMENSIONS_RESOURCE,
+    YOUR_DIMENSIONS_FILTER,
 } from './YourDimensionsPanel/useYourDimensions.js'
 
 const query = {
     dimensions: {
-        resource: yourDimensionsResource,
+        resource: YOUR_DIMENSIONS_RESOURCE,
         params: {
             page: 1,
             pageSize: 1,
             fields: 'id',
-            filter: yourDimensionsFilter,
+            filter: YOUR_DIMENSIONS_FILTER,
         },
     },
 }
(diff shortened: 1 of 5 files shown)


src/components/MainSidebar/YourDimensionsPanel/useYourDimensions.js
```diff
@@ -2,22 +2,25 @@ import { useDataQuery } from '@dhis2/app-runtime'
 import { useEffect, useState } from 'react'
 import { DIMENSION_LIST_FIELDS } from '../DimensionsList/index.js'
 
+const yourDimensionsResource = 'dimensions'
+// Fixed filter on org units for 2.38 ?
+const yourDimensionsFilter = 'dimensionType:eq:ORGANISATION_UNIT_GROUP_SET'
```
HendrikThePendric, line 7: Perhaps we should use `SCREAMING_SNAKE_CASE` for these string constants, as we do elsewhere?

src/components/MainSidebar/YourDimensions.js
```diff
@@ -0,0 +1,42 @@
...
+        params: {
+            page: 1,
+            pageSize: 1,
+            fields: 'id',
+            filter: yourDimensionsFilter,
+        },
+    },
+}
+
+export const YourDimensions = ({ selected, count, onClick }) => {
```
HendrikThePendric, line 24: Perhaps consider renaming to sth like `YourDimensionsMenuItem`?

HendrikThePendric (approved):
Left two small suggestions, but LGTM 💪 .

src/components/MainSidebar/YourDimensionsPanel/useYourDimensions.js
edoardo: Of course.
Fixed.

src/components/MainSidebar/YourDimensions.js
edoardo: Done.

PR #987, opened 2022-02-24
feat: show dashed style on program dimension badge if no program is selected

HendrikThePendric:
Implements KMFT point from [this Slack thread](https://dhis2.slack.com/archives/G3TL0J145/p1645628699820989)

---

### Key features

1. If no program is selected the program dimension menu item count badge now gets a custom dashed style
2. The badge will still display if count equals `0` (before it would not render)
3. I noticed that when a menu item is selected and being hovered, the color of it is very close to the color of the badge, so I added an additional style to remedy this.

---

### Remark

Reg point 3 above, I guess mainly for @cooper-joe:  I think it'd make more sense to go with teal for these badges instead of grey. Firstly because teal is the color we have associated with "selection" in our design system. Secondly because that would fix the issue with the hover-styles and badge styles colliding in a more consistent way.

### Screenshots
*No program selected*
Screenshot 2022-02-24 at 15 39 30

*Program selected*
Screenshot 2022-02-24 at 15 40 14

Note: Both screenshots are showing the hover state of the menu-item

### DIFF
--- a/src/components/MainSidebar/MainSidebar.js
+++ b/src/components/MainSidebar/MainSidebar.js
@@ -7,7 +7,11 @@ import {
     acSetUiAccessoryPanelOpen,
     acSetUiDetailsPanelOpen,
 } from '../../actions/ui.js'
-import { sGetUiInputType, sGetUiShowAccessoryPanel } from '../../reducers/ui.js'
+import {
+    sGetUiInputType,
+    sGetUiShowAccessoryPanel,
+    sGetUiProgramId,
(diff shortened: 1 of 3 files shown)


janhenrikoverland:
@HendrikThePendric Colors should be:

- Default: `background: rgba(160, 173, 186, 0.2);`
- When no program is selected: `border: 1px dashed rgba(74, 87, 104, 0.4); background: transparent;`

edoardo (approved):

PR #990, opened 2022-02-24
feat: apply font size and padding based on settings

HendrikThePendric:
Implements Visualization sizes from options 

---

### Key features

1. Data table cells are padded according to the "Display density" option
2. Data table cells get a font size from the "Font size" option 

---

### Description

This PR comes with some notes:
- The data-table component from @dhis2/ui did already come with a `large` prop, but no `small` prop was available yet, and this is not planned either. Joe advised to apply custom styling in the app, which I did.
- Changing the content of the pagination is a bit tricky. So it was discussed that for the MVP we just decrease the padding a bit when the dispaly density changes. 

---

### TODO

-   [ ] The data table in `@dhis2/ui` uses a slightly different padding and font-size for the header cells than for the body cells. We probably need to do the same for this custom data-table. I've already implemented individual style rules for the header and body cells, but they currently hold the same values. So once @cooper-joe provides the correct values for the header it will be simply a matter of copying these to the stylesheet.
- [ ] Pagination padding values need to be verified by @cooper-joe and adjusted where needed.

---

### Screenshots

*Compact / Small*
Screenshot 2022-02-24 at 16 57 36

*Normal / Normal*
Screenshot 2022-02-24 at 16 59 20

*Comfortable / Large*
Screenshot 2022-02-24 at 17 00 18

*Compact / Large*
Screenshot 2022-02-24 at 17 01 44
*I've included this one because this unexpected combination actually looks pretty useful*

### DIFF
--- a/src/components/Visualization/styles/Visualization.module.css
+++ b/src/components/Visualization/styles/Visualization.module.css
@@ -59,19 +59,63 @@
 .stickyNavigation {
     position: sticky;
     left: 0;
+}
+
+/* Table header cells in various sizes */
+.dataTable .headerCell.fontSizeLarge {
+    font-size: 15px;
+    line-height: 16px;
+}
+.dataTable .headerCell.fontSizeNormal {
+    font-size: 12px;
+    line-height: 13px;
+}
+.dataTable .headerCell.fontSizeSmall {
+    font-size: 10px;
+    line-height: 11px;
+}
+.dataTable .headerCell.sizeComfortable {
+    padding: 16px 12px;
+}
+.dataTable .headerCell.sizeNormal {
+    padding: 8px 8px;
+}
+.dataTable .headerCell.sizeCompact {
+    padding: 4px 8px;
+}
+
+/* Table body cells in various sizes */
+.dataTable .cell.fontSizeLarge {
+    font-size: 15px;
+    line-height: 16px;
+}
+.dataTable .cell.fontSizeNormal {
+    font-size: 12px;
+    line-height: 13px;
+}
+.dataTable .cell.fontSizeSmall {
+    font-size: 10px;
+    line-height: 11px;
+}
+.dataTable .cell.sizeComfortable {
+    padding: 16px 12px;
+}
+.dataTable .cell.sizeNormal {
+    padding: 8px 8px;
+}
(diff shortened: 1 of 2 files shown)


src/components/Visualization/styles/Visualization.module.css
```diff
@@ -59,19 +59,63 @@
...
+}
+.dataTable .headerCell.sizeComfortable {
+    padding: 16px 12px;
+}
+.dataTable .headerCell.sizeNormal {
+    padding: 8px 8px;
+}
+.dataTable .headerCell.sizeCompact {
+    padding: 4px 8px;
 }
```
HendrikThePendric, line 85: Change these values according to input from @cooper-joe

src/components/Visualization/styles/Visualization.module.css
```diff
@@ -59,19 +59,63 @@
...
-    font-size: 10px !important;
+/* Sizes for the table footer */
+.dataTable .stickyNavigation.sizeComfortable {
+    padding: 14px 12px;
+}
+.dataTable .stickyNavigation.sizeNormal {
+    padding: 10px 12px;
+}
+.dataTable .stickyNavigation.sizeCompact {
+    padding: 6px 12px;
```
HendrikThePendric, line 118: Change these values according to input from @cooper-joe

src/components/Visualization/styles/Visualization.module.css
```diff
@@ -59,19 +59,63 @@
...
+    line-height: 11px;
+}
+.dataTable .headerCell.sizeComfortable {
+    padding: 16px 12px;
+}
+.dataTable .headerCell.sizeNormal {
+    padding: 8px 8px;
+}
+.dataTable .headerCell.sizeCompact {
+    padding: 4px 8px;
```
HendrikThePendric, line 84: Change these values according to input from @cooper-joe

cooper-joe:
Updated styles for the header cells, which use slightly different padding to the content cells:

**Padding**
Comfortable: `padding: 12px 12px`,
Normal: `padding: 4px 8px`,
Compact: `padding: 0px 8px`,

**Font size / Line height**
No change, uses same styles as content cells.

janhenrikoverland (approved):

PR #996, opened 2022-02-28
fix: add dimension to metadata when adding from chip menu

HendrikThePendric:
Implements KMFT issue, see [Slack thread](https://dhis2.slack.com/archives/G01PW9YGZD2/p1646042397493249)

---

### Key features

1. Prevents app crashing when adding a dimension to filter/column via the chip menu

---

### Description

This issue was only relevant to dimensions originating from the DB, so only "program" and "your" dimensions. 

---

### DIFF
--- a/src/components/MainSidebar/DimensionItem/DimensionItem.js
+++ b/src/components/MainSidebar/DimensionItem/DimensionItem.js
@@ -59,11 +59,7 @@ export const DimensionItem = ({
     } = useSortable({
         id: draggableId || id,
         disabled: disabled || selected,
-        data: {
-            name,
-            dimensionType,
-            valueType,
-        },
+        data: dimensionMetadata,
     })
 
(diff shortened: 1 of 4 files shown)


src/components/MainSidebar/DimensionItem/DimensionItem.js
```diff
@@ -48,11 +47,7 @@ export const DimensionItem = ({
     } = useSortable({
         id: draggableId || id,
         disabled: disabled || selected,
-        data: {
-            name,
-            dimensionType,
-            valueType,
-        },
+        data: dimensionMetadata,
```
HendrikThePendric, line 50: Note that I also changed this. It's an attempt to standardise which properties are present in the dimension metadata.

janhenrikoverland (approved):

PR #997, opened 2022-02-28
fix: add button for opening Sharing dialog for interpretation

edoardo:
Implements [TECH-1001](https://jira.dhis2.org/browse/TECH-1001)

---

### Key features

1. allow to edit interpretations' sharing settings

---

### Description

This feature is present in the old `d2-ui` implementation.
Together with @janhenrikoverland we decided to remove the sharing access summary for now, the string can easily become too long if we should use the same format used in the About visualization panel.
Only the `Manage sharing` button is shown.

---

### Screenshots

Screenshot 2022-02-28 at 12 08 19


janhenrikoverland (approved):
Approved when comments about button alignment is addressed.

PR #1003, opened 2022-02-28
chore: @dhis2/ui version 8

janhenrikoverland:
Upgrades to @dhis2/ui version 8.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2022-02-28T07:46:52.105Z\n"
-"PO-Revision-Date: 2022-02-28T07:46:52.105Z\n"
+"POT-Creation-Date: 2022-02-28T14:28:34.391Z\n"
+"PO-Revision-Date: 2022-02-28T14:28:34.391Z\n"
 
 msgid "Add to {{axisName}}"
 msgstr "Add to {{axisName}}"
@@ -660,9 +660,6 @@ msgstr "User sub-units"
 
 msgid "User sub-x2-units"
 msgstr "User sub-x2-units"
-
-msgid "This month"
-msgstr "This month"
 
 msgid "Table title"
 msgstr "Table title"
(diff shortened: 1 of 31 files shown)


HendrikThePendric (changes requested):
Major version updates can be problematic because the app-shell is now on a different major version than our app. The result is that yarn will keep two copies around:
```
❯ yarn why @dhis2/ui
yarn why v1.22.17
warning ../../package.json: No license field
[1/4] 🤔  Why do we have the module "@dhis2/ui"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@dhis2/ui@8.0.0"
info Has been hoisted to "@dhis2/ui"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "13.91MB"
info Disk size with unique dependencies: "99.75MB"
info Disk size with transitive dependencies: "120.57MB"
info Number of shared dependencies: 79
=> Found "@dhis2/app-shell#@dhis2/ui@7.16.2"
info This module exists because "@dhis2#cli-app-scripts#@dhis2#app-shell" depends on it.
info Disk size without dependencies: "60KB"
info Disk size with unique dependencies: "85.9MB"
info Disk size with transitive dependencies: "106.72MB"
info Number of shared dependencies: 79
✨  Done in 3.44s.
```

This used to cause problems with the layering system, but we changed the implementation of that and now nothing should break. However, having these two versions around will/could still bloat the bundle size.

I expect in this case we should be able to fix things with a resolutions field.

HendrikThePendric (approved):
In commit [5b15d7b](https://github.com/dhis2/event-reports-app/pull/1003/commits/5b15d7b4823e8f0244be43ae4582296e42eb405a) I added a resolutions fields and I ran `yarn install` again. I think this is now looking better:
```
❯ yarn why @dhis2/ui  
yarn why v1.22.17
warning ../../package.json: No license field
[1/4] 🤔  Why do we have the module "@dhis2/ui"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@dhis2/ui@8.0.0"
info Has been hoisted to "@dhis2/ui"
info Reasons this module exists
   - Specified in "dependencies"
   - Hoisted from "@dhis2#cli-app-scripts#@dhis2#app-shell#@dhis2#ui"
info Disk size without dependencies: "60KB"
info Disk size with unique dependencies: "24.02MB"
info Disk size with transitive dependencies: "44.84MB"
info Number of shared dependencies: 71
✨  Done in 1.39s.
```

PR #1006, opened 2022-03-01
fix: address interpretation modal regressions

HendrikThePendric:
Implements [TECH-1005](https://jira.dhis2.org/browse/TECH-1005)
_And some other issues with the interpretations modal_

---

### Key features

1. Adds whitespace below the title and prevent the characters from being clipped
2. Implement calculated widths for both the modal content and the visualization, to prevent a horizontal scrollbar from appearing
3. Disable interpretation thread when there is a visualization load error

---

### Known issues

-  Both the modal content and the visulaization are using a computed width based on the window width. The `InterpretationsModal` component is going to go into analytics, and the `Visualization` component is staying in the app, or will be a separate analytics component. As such, I didn't see much opportunity for code-sharing right now, so I have a bit of duplicated logic.

---

### Screenshots

*Disabled on load error*
Screenshot 2022-03-01 at 10 54 29

### DIFF
--- a/src/components/Interpretations/InterpretationModal/useModalContentWidth.js
+++ b/src/components/Interpretations/InterpretationModal/useModalContentWidth.js
@@ -1,0 +1,34 @@
+import { useState, useEffect } from 'react'
+import { useDebounce } from '../../../modules/utils.js'
+
+const MODAL_SIDE_PADDING = 2 * 24
+const MODAL_SIDE_MARGINS = 2 * 128
+
+const computeModalContentWidth = (windowWidth) => {
+    return windowWidth - MODAL_SIDE_MARGINS - MODAL_SIDE_PADDING
+}
+
+export const useModalContentWidth = () => {
+    const [windowWidth, setWindowWidth] = useState(window.innerWidth)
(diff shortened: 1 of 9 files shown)


src/components/Interpretations/InterpretationModal/useModalContentWidth.js
```diff
@@ -0,0 +1,34 @@
...
+            window.removeEventListener('resize', onResize)
+        }
+    }, [])
+
+    useEffect(() => {
+        setModalContentWidth(computeModalContentWidth(debouncedWindowWidth))
+    }, [debouncedWindowWidth])
+
+    return modalContentWidth
+}
```
HendrikThePendric, line 34: When I mentioned some code duplication In the PR description, I was referring to the section above.

HendrikThePendric, line 34: When I mentioned some code duplication In the PR description, I was referring to the section above. Some of the constants and the way we week track of the (debounced) window width is identical to what I have in `src/components/Visualization/useAvailableWidth.js`.

edoardo (approved):

PR #1244, opened 2024-01-30
fix: add class-name for push-analytics

HendrikThePendric:
Upgrade d2-analysis and add a class-name directly in the app to allow push-analytics to find the right elements with use text selectors


HendrikThePendric:
Due to a release issue, I ended up resetting the master branch to a commit prior to this and redid these changes. See https://github.com/dhis2/event-reports-app/pull/1245#issue-2107382367  for details.

